PR #13, opened 2024-03-15
Fix starvation with async server and interleaving optimization

JoeZijunZhou:
- Make server async to handle high throughput from client (the number of requests a sync server can handle is limited by server side thread number, since each thread handles a request in blocking manner).
- Add AsyncMultiFuture data structure for the retuern_channel to stream response tokens asynchronously (threadsafe ensures no token drop in async server)
- Using blocking fixed size queue to block and yield threads efficiently
  - blocking for get and put operations for generate and detokenize queue
  - blocking for prefill queue get operation
  - unblocking for prefill queue put operation (unbound queue size to enqueue all requests from client)
- Optimized interleaving prefill, insert, and generate
  - When decode slots are empty, yield to prefill, then insert the prefill results as many as possible to the decode slots
  - When decode slots are full, block insert and prefill, keep decoding and detokenizing until some deocde completes and returns available slot.
  - generate queue size set to 3/1 decode batch size
    - get decode slots saturated fast (larger generate queue size ensures it has enough requests to insert to slots)
    - avoid OOM when both decode slots and generate queue are both saturated (can't set generate queue size too large, otherwise too many prefill results enqueued in generate queue)

### DIFF
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -85,12 +85,13 @@ import sys
 import threading
 import time
 import traceback
-from typing import Any, Iterable, Optional, Union
+from typing import Any, AsyncIterator, Optional, Union
 
 import grpc
 import jax
 from jetstream.core.proto import jetstream_pb2
 from jetstream.core.proto import jetstream_pb2_grpc
+from jetstream.core.utils import async_multifuture
 from jetstream.engine import engine_api
 from jetstream.engine import token_utils
 import numpy as np
@@ -136,13 +137,7 @@ class ActiveRequest:
   # We keep prefill and decode information together in the same object so that
   # there is less indirection about where this return channel is.
   # The return channel returns a list of strings, one per sample for that query.
-  return_channel: queue.Queue[list[str]] = dataclasses.field(
-      default_factory=queue.Queue
-  )
-
-  def next(self: 'ActiveRequest'):
-    """Blocks until the next token is available, call from RPC threads."""
-    return self.return_channel.get()
+  return_channel: async_multifuture.AsyncMultifuture[list[str]] = None
 
   def enqueue_tokens(self, generated_tokens: list[str]):
     """Records information about the step.
@@ -152,7 +147,7 @@ class ActiveRequest:
 
     This should be called only from within the Drivers background thread.
     """
-    self.return_channel.put(generated_tokens)
+    self.return_channel.add_result(generated_tokens)
 
 
 class JetThread(threading.Thread):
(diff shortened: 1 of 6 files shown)


qihqi (approved):

rwitten:
This CR cannot merge. It is driving contention (probably GIL) causing stalls.
https://xprof.corp.google.com/trace_viewer/rwitten-7572422995686954905?host_index=0

rwitten (changes requested):
We cannot merge these changes because they drive a duty cycle migration.

jetstream/core/utils/async_multifuture.py
```diff
@@ -0,0 +1,89 @@
+import asyncio
```
vipannalla, line 1: Where is this module used? I can't find any references.

vipannalla:
Can you please add a xprof with the new changes so we can compare against the GIL-contention xprof that Rafi provided and rule-out any contention issues?

jetstream/core/utils/async_multifuture.py
JoeZijunZhou: In orchestrator, the `return_channel`, https://github.com/google/JetStream/pull/13/files/09e0e659ba821e14c5595dbfb3c8ec620725d865#diff-753df909960695fd6aa61b95c4762c4edc50a298007d40de7096a7e49e0e1215R130

JoeZijunZhou:
> Can you please add a xprof with the new changes so we can compare against the GIL-contention xprof that Rafi provided and rule-out any contention issues?

Added a xprof in https://docs.google.com/document/d/1PWje3lup0ZHjtgbcWQbK8r1nNwgKE2rHBdVLbWU2aas/edit?tab=t.0#heading=h.f39bxdhl45e2 (Since here is public, just shared in the internal doc).

jetstream/core/orchestrator.py
```diff
@@ -242,64 +254,62 @@ def __init__(
...
-    #   operation. I.e. potentially available rows in the batch and/or
-    #   microbatch. When we want to insert a prefill result, we pop an integer
-    #   to insert at. When this is empty, it means all slots are full.
-
-    # Construct a)
-    self._generate_slots = [queue.Queue() for _ in generate_engines]
+    self._detokenize_backlogs = [
+        # We don't let detokenization accumulate more than 8 steps to avoid
+        # synchronization issues.
+        queue.Queue(8)
```
vipannalla, line 266: How did you comeup with 8? Is that hardware dependent -- will this work for all configs (v5e-8, v5e-1 etc?)

vipannalla (commented):
Thanks for stacktrace, fix looks good. I had a quick question about detokenize queue

jetstream/core/orchestrator.py
JoeZijunZhou: The JET has an investigation on GIL contention issue and detokenize queue size, and figured out that this size can avoid synchronization issue. It's not dependent to hardware.

jetstream/core/orchestrator.py
```diff
@@ -469,15 +469,46 @@ def _generate_thread(
...
+          else:
+            break
+
+        # Signal to kill the thread.
+        if new_request is None:
+          return
+
         logging.info(
-            'Generate slice %d slot %d step %d',
+            'Generate slice %d filling slot %d at step %d.',
```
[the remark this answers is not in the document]
JoeZijunZhou: Agree, added a TODO above the insert while loop.

zhihaoshan-google (approved):

PR #16, opened 2024-03-22
Enable jax profiler server

morgandu:


.gitignore
```diff
@@ -2,4 +2,8 @@ __pycache__
 .env*
 build/
 dist/
-google_jetstream.egg-info/
\ No newline at end of file
+google_jetstream.egg-info/
+
+tmp/
+venv/
+logs/
```
gangji, line 9: Can we keep it sorted? Thanks!

vipannalla:
Were you able to successfully generate profile on-demand profile? Might be good to add steps to use profiler here and/or in README

jetstream/core/server_lib.py
```diff
@@ -94,6 +94,10 @@ def run(
 
   jetstream_server = JetStreamServer(driver, server)
   jetstream_server.start(port, credentials)
+
+  jax.profiler.start_server(9999)
```
JoeZijunZhou, line 98: Let's add the profiler server in the server implementation with a flag to control that? https://github.com/google/JetStream/blob/main/jetstream/core/implementations/mock/server.py
Since we only need it during debugging and profiling, and start an extra server may affect JetStream server's performance.

PR #21, opened 2024-03-29
Update benchmark to run openorca dataset

morgandu:


FanhaiLu1 (approved):

JoeZijunZhou (approved):

JoeZijunZhou (commented):
We can merge this change first, since I need to release a new JetStream py package. We can do refactor later since I see the current sample filter logic is identical for both dataset.

benchmarks/benchmark_serving.py
```diff
@@ -167,27 +161,104 @@ def sample_requests(
...
+  # Load the dataset.
+  with open(dataset_path) as f:
+    dataset = json.load(f)
+
+  # Tokenize the prompts and completions.
+  prompts = dataset["prompts"]
+  outputs = dataset["results"]
+  n = len(prompts)
+  prompt_token_ids = tokenizer.tokenize(prompts)
+  output_token_ids = tokenizer.tokenize(outputs)
```
JoeZijunZhou, line 211: I feel we could extract this part out as a func for different dataset, and the rest are identical and thus we could keep them in the sample_request function?

morgandu: Discussed offline, there are some existing data processing that may not be necessary. Will revisit and refactor the data preprocessing part if needed.

PR #22, opened 2024-03-29
Fix llama2 decode bug in tokenizer

FanhaiLu1:
This PR fix llama2 decode bug in tokenizer. There 344 token ids are decoded to hex str before this PR, all these 344 token ids are decoded to right str value with this PR. 

Below are decode results for token_id 13 (before vs with this PR):

-  <0x0A>

-  \n 

Here is an example of output with this PR: 

>  to find purpose and fulfillment. Here are some reasons why:\n\n1. Purpose gives life meaning: Ha
> ving a purpose in life gives it meaning and direction. It helps us to focus on what is important and to make decisions that 
> align with our values and goals.\n2. Fulfillment comes from pursuing purpose: When we pursue our purpose, we experience fulf
> illment and satisfaction. This is because we are doing something that is meaningful and important to us, and we are making p
> rogress towards our goals.\n3. Purpose is unique to each person: Everyone has their own unique purpose in life, and it is im
> portant to discover and pursue our own purpose. This is because when we are doing something that is meaningful and important
>  to us, we are more likely to be happy and fulfilled.\n4. Purpose can be found in many areas of life: Our purpose can be fou
> nd in many areas of life, such as our work, relationships, hobbies, and personal growth. It is important to find ways to inc
> orporate our purpose into all of these areas of life.\n5.


jetstream/engine/token_utils.py
```diff
@@ -28,6 +28,19 @@
...
 
 
+def mix_decode(vocab: Vocabulary, tok_id: int):
+    """
+    The IdToPiece and decode results differ for 344 tokens in Llama2. 
+    Use the decode function to generate the correct strings for these 344 tokens. 
+    If IdToPiece returns a hex string (e.g., '<0x0A>') for a token within these 344,
+      utilize IdToPiece to convert it into a string, likely with a space placeholder (' ') for the corresponding tokens.
+    """
+    p_token = vocab.tokenizer.IdToPiece([tok_id])
```
patemotter, line 38: It looks like tests are failing because there is a discrepancy between the the real `IdToPiece` and mocked version expecting different parameter types. `int` vs `[int]`.

jetstream/engine/token_utils.py
```diff
@@ -28,6 +28,19 @@
...
+    p_token = p_token[0].replace('▁', ' ').replace('_', ' ')
```
patemotter, line 39: Nit: You might want a clarifying comment here because it looks like these two characters are the same since the difference is so subtle. It looks like you had that comment originally but removed it in this change.

jetstream/tests/engine/test_token_utils.py
```diff
@@ -55,8 +52,31 @@ def test_decode(self):
...
+       self.assertNotEqual(jt_output, expeted_sp_output)   
+
+
+    def test_mix_decode(self):
+       self.setup()
+       for n in range(0, self.sp_tokenizer.tokenizer.vocab_size()):
+          #decode
+          d_t = self.sp_tokenizer.decode([n])
+          #IdToPiece
+          p_t = self.jt_tokenizer.decode([n])
```
patemotter, line 64: Nit: rather than comments here use more descriptive variable names.

jetstream/engine/token_utils.py
FanhaiLu1: Thanks Pate! The are some weird type error when I pass int to IdToPiece, I changed to int[] for type checking, the int type is right one.  Fixed it now.

jetstream/engine/token_utils.py
FanhaiLu1: Yes, current logic of space is not correct. I first made change in this PR, but then I planed to do it in a new PR as this PR focus on the hex str part. Below are some token token impacted examples FYI: 
903;_;  ; _
['▁_']
4770;__;   ; __
['▁__']
7471;$_; $ ; $_
['▁$_']
7722;_{;  {; _{
['▁_{']
9423;(_; ( ; (_
['▁(_']
11119;"_; " ; "_
['▁"_']
17117;_,;  ,; _,
['▁_,']
19392;`_; ` ; `_
['▁`_']
21326;`__; `  ; `__
['▁`__']
22868;'_; ' ; '_
["▁'_"]
23160;[_; [ ; [_
['▁[_']
24459;_);  ); _)
['▁_)']

jetstream/tests/engine/test_token_utils.py
FanhaiLu1: Changed, thanks Pate!

patemotter (approved):
LGTM

PR #25, opened 2024-04-01
Update ckpt conversion scripts

JoeZijunZhou:
- Remove the base ckpt copy to GCS bucket step from the script, let users copy the ckpt themselves and pass the `$CHKPT_BUCKET` value as an argument of the script.


vipannalla (approved):
LGTM

PR #27, opened 2024-04-03
move tokenizer model to third party llama2

FanhaiLu1:
Move tokenizer model to third party llama2


jetstream/tests/engine/test_token_utils.py
```diff
@@ -48,7 +48,7 @@ def decode(self, t: int) -> str:
 
 class TokenUtilsTest(unittest.TestCase):
     def setup(self):
-        tokenizer_path = "tokenizer.model"
+        tokenizer_path = "third_party/llama2/tokenizer.model"
         current_dir = os.path.dirname(__file__)
         tokenizer_path = os.path.join(current_dir, tokenizer_path)
         print(f"model_path: {tokenizer_path}")
```
patemotter, line 54: Nit outside of this change: Should probably be renamed to tokenizer_path. 

Also typos in comments above "Tokenier" -> "Tokenizer"

patemotter (approved):
Overall LGTM. Would some of these tests break if we were to change the tokenizer being used?

jetstream/tests/engine/test_token_utils.py
FanhaiLu1: Thanks Pate! there could be more data set from llama2. Keep it as "third_party/llama2"

FanhaiLu1:
> Overall LGTM. Would some of these tests break if we were to change the tokenizer being used?

The test_underscore_in_output test is specific for a  under score token id, it may fail if tokenizer changed. All other 3 test are  tokenizer independent

PR #32, opened 2024-04-09
add sample_idx in InputRequest for debugging

morgandu:


benchmarks/benchmark_serving.py
```diff
@@ -267,9 +283,13 @@ def sample_requests(
 
   print(f"{len(sampled_indices)=}")
   # some of these will be filtered out, so sample more than we need
-  dataset = [dataset[i] for i in sampled_indices]
 
-  tokenized_dataset = tokenize_dataset(dataset, tokenizer)
+  sampled_dataset = []
```
gangji, line 287: Just curious, numpy has `numpy.take`?

benchmarks/benchmark_serving.py
```diff
@@ -98,6 +98,7 @@ class InputRequest:
   prompt_len: int = 0
   output: str = ""
   output_len: int = 0
+  sample_idx: int = -1
```
FanhaiLu1, line 101: There is right mapping between request and output in benchmark. what is the purpose of sample_idx? Will it help you find the related request easily?

FanhaiLu1 (approved):

benchmarks/benchmark_serving.py
vipannalla: +1, your goal is to be able to print the input prompt for the output for debugging right?

morgandu: My goal is to be able to locate the original order / index from the original dataset file. 

Say, if there are 10k data samples in the original dataset file, after randomly sampling, being passed as input request into server, the requests are returned based on decode completions. Currently we save the prompt, original result, and generated result in the request output file. If there are other metadata I am interested in checking in the original dataset file. How can I locate them?

FanhaiLu1: agree, it's hard to locate original dateset. Please also feel free to add the important metadata into the result.

PR #49, opened 2024-04-23
Add np padded token support

FanhaiLu1:
Current code only return jax padded token, this PR added np padded token support. It bring two benefits: 

1: Frame framework independent. Not depend on Jax framework 
2: Serialization benefits

The change will not impact current behaviors, existing code can still use Jax padded tokens without any change.

### DIFF
--- a/jetstream/engine/token_utils.py
+++ b/jetstream/engine/token_utils.py
@@ -16,7 +16,7 @@
 
 from bisect import bisect_left
 import logging
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Union
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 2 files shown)


vipannalla:
Looks good to me. Can you please add/update tests for the new padding option in [test_token_utils.py](https://github.com/google/JetStream/blob/main/jetstream/tests/engine/test_token_utils.py)

vipannalla (approved):
Awesome!

FanhaiLu1:
> Looks good to me. Can you please add/update tests for the new padding option in [test_token_utils.py](https://github.com/google/JetStream/blob/main/jetstream/tests/engine/test_token_utils.py)

Thanks, added.

PR #53, opened 2024-04-24
Add an abstract class for Tokenizer

bhavya01:


bhavya01:
I see that #40 changed the implementation to output only token IDs instead of strings.  Will need to update this PR with the new changes.

jetstream/engine/engine_api.py
```diff
@@ -200,7 +253,15 @@ def get_prefix_destination_sharding(self) -> Any:
...
-    """Returns the info to construct a sentencepiece tokenizer in py/c++."""
+    """Returns the info to construct a tokenizer in py/c++."""
+
+  # pylint: disable=unused-argument
+  def build_tokenizer(
+      self,
+      metadata: tokenizer_pb2.TokenizerParameters,
+  ) -> Tokenizer:
+    """Builds a new tokenizer object and returns it."""
+    return None
```
qihqi, line 264: maybe add a default impl here to make it not BC breaking

bhavya01: Did this to avoid circular import issue. In orchestrator.py, if tokenizer is None, we build the default SentencePieceTokenizer

jetstream/engine/engine_api.py
```diff
@@ -124,6 +124,60 @@ def get_result_at_slot(self, slot: int) -> SlotData:
     )
 
 
+class Tokenizer(abc.ABC):
+  """Tokenizer to convert strings to token ids and vice-versa."""
+
+  @abc.abstractmethod
+  def encode(self, s: str, **kwargs):
```
JoeZijunZhou, line 131: Could we add return types for encode?

jetstream/engine/token_utils.py
```diff
@@ -196,3 +197,79 @@ def load_vocab(path: str, extra_ids: int = 0) -> Vocabulary:
...
     return vocab
+
+
+class SentencePieceTokenizer(engine_api.Tokenizer):
+  """Tokenizer to convert strings to token ids and vice-versa."""
+
+  def __init__(self, metadata: tokenizer_pb2.TokenizerParameters):
+    self.vocab = load_vocab(metadata.path, metadata.extra_ids)
+
+  def encode(self, s: str, **kwargs):
```
JoeZijunZhou, line 208: Could we add return types for encode?

JoeZijunZhou (approved):
Could we add return types for encode?

jetstream/engine/engine_api.py
```diff
@@ -19,7 +19,7 @@
 """
 
 import abc
-from typing import Any, Optional, Tuple, Union
+from typing import Any, Optional, Tuple, Union, List
```
JoeZijunZhou, line 22: Maybe use list instead of List?

jetstream/engine/engine_api.py
```diff
@@ -124,6 +124,60 @@ def get_result_at_slot(self, slot: int) -> SlotData:
     )
 
 
+class Tokenizer(abc.ABC):
```
FanhaiLu1, line 127: A  build_tokenizer function is sufficient for engine_api.py.  Could you move Tokenizer out of engine_api.py and create a class tokenizer_ap.py or related class to store Tokenizer?

jetstream/core/orchestrator.py
```diff
@@ -568,8 +569,9 @@ def _detokenize_thread(self, idx: int):
     my_slots = self._generate_slots[idx]
 
     metadata = my_generate_engine.get_tokenizer()
-    vocab = token_utils.load_vocab(metadata.path, metadata.extra_ids)
-
+    tokenizer = my_generate_engine.build_tokenizer(metadata)
+    if not tokenizer:
```
FanhaiLu1, line 573: If we let build_tokenizer handle build tokenizer, lest's give engine full responsibility of build_tokenizer.  Engine implementation call token_utils.SentencePieceTokenizer if needed. The error are expected if tokenizer is still None after build_tokenizer. 

Same as line 400 to 402

FanhaiLu1 (changes requested):

jetstream/core/orchestrator.py
bhavya01: I agree with your point but changing the build_tokenizer default implementation to return token_utils.SentencePieceTokenizer will cause a circular import issue
engine_api -> token_utils -> engine_api. For now, I think it would be better to keep it this way and then I can update the MaxText engine to return a SentencePiece Tokenizer.

bhavya01: Updated it remove the circular import by using Any typing annotation.

FanhaiLu1: Great! Thanks for the change and fixing the circular import issues!

FanhaiLu1 (approved):
Thanks, all looks good to me now.

vipannalla:
Thanks for the PR, this looks great. I'd appreciate if you can also add unittests (or update existing ones) for the changes.

PR #67, opened 2024-05-01
Support llama3 tokenizer

bhavya01:


vipannalla (approved):
Looks good.

jetstream/engine/token_utils.py
```diff
@@ -40,9 +41,10 @@ def take_nearest_length(lengths: list[int], length: int) -> int:
   return lengths[pos]
 
 
-def tokenize_and_pad(
```
FanhaiLu1, line 43: Can we still keep this function?

bhavya01: Any particular reason to keep the function?

I changed it to pad_tokens because different tokenizers use different function calls to tokenize the strings. It's easier to have a function just for padding and let each tokenizer convert strings to token IDs.

FanhaiLu1: I do agree separate two part will be clear, meanwhile it bring two functions instead of just one functions to tokenizer and pad. Engine implementation still use this tokenize_and_pad (i.e. interactive mode from pytorch side), it easier for client if the API is consistency from different versions. I feel it would be better to keep both of them and let client decide which one they would like to use. In general, this is a nit comments and won't block your current PR, some one can change this later.

FanhaiLu1 (approved):

jetstream/engine/token_utils.py
bhavya01: Thank you. I have a PR to update the pytorch interactive mode. So, that should fix it.

PR #71, opened 2024-05-03
Prometheus Metrics

Bslabe123:
Added a Gauge for watching the prefill backlog size. Note port 9090 is conventional for prometheus metrics.


vipannalla:
Looks good. Can we do a end-2-end run to ensure there no impact on perf? 
+ Zijun, PTAL .

FanhaiLu1 (commented):
Could you share the Prometheus Metrics results?

jetstream/core/orchestrator.py
```diff
@@ -442,6 +449,8 @@ def _prefill_thread(self, idx: int):
       my_transfer_backlog = self._transfer_backlogs[idx]
       # The prefill thread can just sleep until it has work to do.
       request = self._prefill_backlog.get(block=True)
+      self._prefill_backlog_size_metric.set(self._prefill_backlog.qsize())
```
liurupeng, line 452: qq, should we record the metrics in to places?

liurupeng:
> Could you share the Prometheus Metrics results?

@Bslabe123 yep, it would be good if we could verify the metric is added in the container logs, thanks!

jetstream/core/orchestrator.py
liurupeng: I would assume this is the real prefill queue size during runtime @FanhaiLu1 right?

jetstream/core/orchestrator.py
```diff
@@ -421,6 +427,7 @@ def place_request_on_prefill_queue(self, request: ActiveRequest):
     """Used to place new requests for prefilling and generation."""
```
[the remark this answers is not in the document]
JoeZijunZhou: This is adding requests from JetStream's client to JetStream Orchestrator's prefill backlog. It doesn't relate to engines.

jetstream/core/orchestrator.py
JoeZijunZhou: L430 logs the prefill backlog queue size after a request added to the queue; L452 logs the prefill backlog queue size after a request remove/get from the queue.

JoeZijunZhou (commented):
LGTM! Would also like to see the E2E validation result of this metric.

jetstream/core/orchestrator.py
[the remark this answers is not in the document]
liurupeng: I see, but where is the API used? I didn't find the usage... so I'm not sure whether we should record metrics here

jetstream/core/orchestrator.py
liurupeng: then I would assume we only need to add it in one place right?

JoeZijunZhou: IIUC, @Bslabe123 is trying to collect the prefill queue size metric as the first step?

jetstream/core/orchestrator.py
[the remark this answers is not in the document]
JoeZijunZhou: Which API? This is part of the workflow of the JetStream Decode API.

jetstream/core/orchestrator.py
```diff
@@ -242,6 +246,9 @@ def __init__(
     # Stage 1
     # At first, a request is placed here in order to get prefilled.
     self._prefill_backlog = queue.Queue()
+    self._prefill_backlog_size_metric = prometheus_client.Gauge(
```
JoeZijunZhou, line 249: You can get the decode slots size at line L556 and L587. @liurupeng @Bslabe123 It's similar to the prefill backlog metric.

Bslabe123:
> > Could you share the Prometheus Metrics results?
> 
> @Bslabe123 yep, it would be good if we could verify the metric is added in the container logs, thanks!

Able to validate via also deploying a prometheus server to the cluster to scrape the newly emitted metrics, will follow up with an [ai-on-gke](https://github.com/GoogleCloudPlatform/ai-on-gke) PR with validation instructions/demo

Bslabe123:
> LGTM! Would also like to see the E2E validation result of this metric.

Deployed on GKE with this [maxtext setup](https://github.com/GoogleCloudPlatform/ai-on-gke/tree/main/tutorials-and-examples/inference-servers/jetstream/maxtext/single-host-inference) plus a prometheus server, these are metrics after  1000 concurrent `curl`s to the maxtext http endpoint via `seq 100 | xargs -P 100 -n 1 curl --request POST --header "Content-type: application/json" -s localhost:8000/generate --data '{
    "max_tokens": 200
}'`
 
Screenshot 2024-05-09 at 10 35 55 AM

Bslabe123:
> Looks good. Can we do a end-2-end run to ensure there no impact on perf?
> 
> * Zijun, PTAL .

Timed the above 100 requests with  `bash`s `time` and got the following
```
real	0m26.914s
user	0m0.940s
sys	0m1.313s
```
Can do more extensive benchmarking if these numbers seem questionable

liurupeng:
lgtm, thanks @Bslabe123 !

FanhaiLu1 (approved):
Is Prometheus Metrics mandatory?  If the port already been used or something wrong crash  Prometheus server, what would be impact to jet stream sever? In current logic, it would also crash jetstream sever. 

Can you add flag and only enable  Prometheus Metrics sever and collection if the flag is enabled?

JoeZijunZhou (approved):

PR #77, opened 2024-05-07
Add metadata metrics

yeandy:
Add the ability to pass in additional metadata to be added to the metrics json. For example, adding hardware-related metrics (ici_fsdp_parallelism, ici_autoregressive_parallelism, ici_tensor_parallelism) will enable us to drill down performance by how we shard the model.

### DIFF
--- a/benchmarks/benchmark_serving.py
+++ b/benchmarks/benchmark_serving.py
@@ -612,9 +612,9 @@ def main(args: argparse.Namespace):
     )
 
     metrics_json = {
-      **metrics_json,
-      **benchmark_result,
-      **json.loads(args.additional_metadata_metrics_to_save)
+        **metrics_json,
+        **benchmark_result,
+        **json.loads(args.additional_metadata_metrics_to_save),
     }
     if args.run_eval:
       metrics_json = {**metrics_json, **eval_json}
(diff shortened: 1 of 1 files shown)


benchmarks/benchmark_serving.py
```diff
@@ -611,7 +611,11 @@ def main(args: argparse.Namespace):
         args.request_rate if args.request_rate < float("inf") else "inf"
     )
 
-    metrics_json = {**metrics_json, **benchmark_result}
+    metrics_json = {
+        **metrics_json,
+        **benchmark_result,
+        **json.loads(args.additional_metadata_metrics_to_save),
```
JoeZijunZhou, line 617: Should this metadata be appended to dimensions_json instead of metrics_json? Since metrics values are numerical, mainly for benchmark results.

yeandy: The metadata name is misleading. The values I've been adding are numerical right now. I could rename it from "metadata" to something else?

JoeZijunZhou: These `ici_fsdp_parallelism, ici_autoregressive_parallelism, ici_tensor_parallelism` are you referring to the maxtext config? These are metadata but not metrics?

yeandy: Shared with you an explanation of the metadata via chat.

yeandy: Per our chats, will stringify the ici values, and pass it into the `dimensions_json`.

JoeZijunZhou (approved):
LGTM!

PR #78, opened 2024-05-07
Update JetStream grpc proto to support I/O with text and token ids

JoeZijunZhou:
- Update gprc proto to support
  - Request: token id or text (one of).
  - Response: token id, text or both of them.
  - When input token id, return only token ids. (client tokenization (MLPerf) mode)
  - When input text, return both text and token ids.
- Refactored detokenization handling and Tokenizer API to decouple the logics.
- Added complete support for SentencePiece tokenizer streaming decoding (ensured output text correctness).
- Added and update unit tests for token utils, orchestrator, and server.
- Benchmark script client-side tokenization is enabled.

### DIFF
--- a/jetstream/core/proto/jetstream.proto
+++ b/jetstream/core/proto/jetstream.proto
@@ -19,14 +19,13 @@ package jetstream_proto;
 // TODO: Merge this with main JetStream core once we settle on an API.
 
 service Orchestrator {
-  // Generate the next model tokens.
+  // Query LLM to generate text or tokens.
   rpc Decode(DecodeRequest) returns (stream DecodeResponse) {}
 }
+
 message DecodeRequest {
   // Where to load any pre-existing kv cache from.
   string session_cache = 1;
-  // New text from a user or tool.
-  string additional_text = 2;
   int32 priority = 3;
   // The maximum output length of a sequence. It's used in JetStream to control
   // the output/decode length of a sequence. It would not be used in the engine.
@@ -35,12 +34,44 @@ message DecodeRequest {
   // sequence; max_prefill_predict_length is the maximum length of the
   // input/prefill of a sequence.
   int32 max_tokens = 4;
+
+  message TextContent {
+    string text = 1;
+  }
+  message TokenContent {
+    repeated int32 token_ids = 1;
+  }
+
+  // The client can pass the inputs either as a string, in which case the server will
+  // tokenize it, or as tokens, in which case it's the client's responsibility to
+  // ensure they tokenize its input strings with the correct tokenizer.
+  oneof content {
+    TextContent text_content = 5;
+    TokenContent token_content = 6;
+  }
+  reserved 2;
+  // Next ID: 7
 }
+
 message DecodeResponse {
-  // List of responses, one per sample. The list size depends on text generation strategy the engine used.
-  repeated RepeatedTokenIds response = 1;
-}
-message RepeatedTokenIds {
-  // List of token ids, one list per sample. When speculative decoding is disabled, the list size should be 1; When speculative decoding is enabled, the list size should be >= 1.
-  repeated int32 token_ids = 1;
+  // InitialContent supports returning initial one-off response data from the
+  // stream. It's a placeholder for future features such as history cache.
+  message InitialContent {}
+  message StreamContent {
+    message Sample {
+      // The text string decoded from token id(s).
+      string text = 1;
+      // List of token ids, one list per sample. When speculative decoding is disabled, the list size should be 1; When speculative decoding is enabled, the list size should be >= 1.
+      repeated int32 token_ids = 2;
+    }
+    // Supports multiple samples in the StreamContent. The Sample list size depends on text generation strategy the engine used.
+    repeated Sample samples = 1;
+  }
+
+  oneof content {
+    InitialContent initial_content = 2;
+    StreamContent stream_content = 3;
+  }
+  reserved 1;
+  // Next ID: 4
 }
--- a/benchmarks/benchmark_serving.py
+++ b/benchmarks/benchmark_serving.py
@@ -555,6 +555,7 @@ def main(args: argparse.Namespace):
         args.total_mock_requests
     )  # e.g. [("AB", 2, "AB", 3)]
   else:
+    dataset = []
     if args.dataset == "openorca":
       dataset = load_openorca_dataset_pkl()
     elif args.dataset == "sharegpt":
--- a/benchmarks/benchmark_serving.py
+++ b/benchmarks/benchmark_serving.py
@@ -388,10 +388,10 @@ async def grpc_async_request(
     token_list = []
     request_start_time = time.perf_counter()
     response = stub.Decode(request)
-    async for sample_list in response:
+    async for resp in response:
       if ttft == 0:
         ttft = time.perf_counter() - request_start_time
-      token_list.extend(sample_list.response[0].token_ids)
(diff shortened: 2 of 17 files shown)


FanhaiLu1 (commented):
Any reason to add text back, I suggested we keep both str and id in response in https://github.com/google/JetStream/pull/40. The answer is " don't want to decode it to str (or piece) in jetstream". Would you like to share the reasons to do make the total different decision in just two weeks?  

Below are comment from PR 40: 


> @[FanhaiLu1](https://github.com/FanhaiLu1) FanhaiLu1 [2 weeks ago]
> Can we still keep a str as option? The internal keep both text and token id.
> 
> @[JoeZijunZhou] [2 weeks ago]
> I guess we don't want to decode it to str (or piece) in jetstream, since it would have some off in the final result.

JoeZijunZhou:
> Any reason to add text back, I suggested we keep both str and id in response in #40. The answer is " don't want to decode it to str (or piece) in jetstream". Would you like to share the reasons to do make the total different decision in just two weeks?
> 
> Below are comment from PR 40:
> 
> > @[FanhaiLu1](https://github.com/FanhaiLu1) FanhaiLu1 [2 weeks ago]
> > Can we still keep a str as option? The internal keep both text and token id.
> > @[JoeZijunZhou] [2 weeks ago]
> > I guess we don't want to decode it to str (or piece) in jetstream, since it would have some off in the final result.

Mainly for our customer's request, they would like to have server-side detokenization support. This PR will include the complete support. " don't want to decode it to str (or piece) in jetstream" is due to the incorrect streaming detokenization previously.

jetstream/core/proto/jetstream.proto
```diff
@@ -35,12 +34,44 @@ message DecodeRequest {
...
+    string text = 1;
+  }
+  message TokenContent {
+    repeated int32 token_ids = 1;
+  }
+
+  // The client can pass the inputs either as a string, in which case the server will
+  // tokenize it, or as tokens, in which case it's the client's responsibility to
+  // ensure they tokenize its input strings with the correct tokenizer.
+  oneof content {
```
FanhaiLu1, line 48: It could be both. Client could have 3 choice: token_id, token_text or both of them. If we only support one of them, it's high possible that we need to refactor the code one more time.

FanhaiLu1:
> > Any reason to add text back, I suggested we keep both str and id in response in #40. The answer is " don't want to decode it to str (or piece) in jetstream". Would you like to share the reasons to do make the total different decision in just two weeks?
> > Below are comment from PR 40:
> > > @[FanhaiLu1](https://github.com/FanhaiLu1) FanhaiLu1 [2 weeks ago]
> > > Can we still keep a str as option? The internal keep both text and token id.
> > > @[JoeZijunZhou] [2 weeks ago]
> > > I guess we don't want to decode it to str (or piece) in jetstream, since it would have some off in the final result.
> 
> Mainly for our customer's request, they would like to have server-side detokenization support. This PR will include the complete support. " don't want to decode it to str (or piece) in jetstream" is due to the incorrect streaming detokenization previously.

Thanks, can you create a issues and add the customer's request details? It will be good context for readers.

jetstream/core/proto/jetstream.proto
JoeZijunZhou: This is the request input, user should only input text or token. The response can return token_id, token_text or both of them. The control logic will be implemented in the orchestrator.

JoeZijunZhou:
Addressing #79 .

FanhaiLu1:
> > Any reason to add text back, I suggested we keep both str and id in response in #40. The answer is " don't want to decode it to str (or piece) in jetstream". Would you like to share the reasons to do make the total different decision in just two weeks?
> > Below are comment from PR 40:
> > > @[FanhaiLu1](https://github.com/FanhaiLu1) FanhaiLu1 [2 weeks ago]
> > > Can we still keep a str as option? The internal keep both text and token id.
> > > @[JoeZijunZhou] [2 weeks ago]
> > > I guess we don't want to decode it to str (or piece) in jetstream, since it would have some off in the final result.
> 
> Mainly for our customer's request, they would like to have server-side detokenization support. This PR will include the complete support. " don't want to decode it to str (or piece) in jetstream" is due to the incorrect streaming detokenization previously.

Soon, there could be another customer ask both token id an text support? I feel we should predict what are the customer's requirements and do it in once. Support token id, text and both is one solution for all the customers.

JoeZijunZhou:
> > > Any reason to add text back, I suggested we keep both str and id in response in #40. The answer is " don't want to decode it to str (or piece) in jetstream". Would you like to share the reasons to do make the total different decision in just two weeks?
> > > Below are comment from PR 40:
> > > > @[FanhaiLu1](https://github.com/FanhaiLu1) FanhaiLu1 [2 weeks ago]
> > > > Can we still keep a str as option? The internal keep both text and token id.
> > > > @[JoeZijunZhou] [2 weeks ago]
> > > > I guess we don't want to decode it to str (or piece) in jetstream, since it would have some off in the final result.
> > 
> > 
> > Mainly for our customer's request, they would like to have server-side detokenization support. This PR will include the complete support. " don't want to decode it to str (or piece) in jetstream" is due to the incorrect streaming detokenization previously.
> 
> Soon, there could be another customer ask both token id an text support? I feel we should predict what are the customer's requirements and do it in once. Support token id, text and both is one solution for all the customers.

This already supports token id, text and both as response. User will not input both text and tokens at the same in 1 request right? Why will user input both text and tokens to request JetStream API? 

I was thinking simplified it to 2 mode:
1. For MLPerf, return token ids only
2. Other cases, always return text + token ids + score (to be added to Sample)

FanhaiLu1:
> > > > Any reason to add text back, I suggested we keep both str and id in response in #40. The answer is " don't want to decode it to str (or piece) in jetstream". Would you like to share the reasons to do make the total different decision in just two weeks?
> > > > Below are comment from PR 40:
> > > > > @[FanhaiLu1](https://github.com/FanhaiLu1) FanhaiLu1 [2 weeks ago]
> > > > > Can we still keep a str as option? The internal keep both text and token id.
> > > > > @[JoeZijunZhou] [2 weeks ago]
> > > > > I guess we don't want to decode it to str (or piece) in jetstream, since it would have some off in the final result.
> > > 
> > > 
> > > Mainly for our customer's request, they would like to have server-side detokenization support. This PR will include the complete support. " don't want to decode it to str (or piece) in jetstream" is due to the incorrect streaming detokenization previously.
> > 
> > 
> > Soon, there could be another customer ask both token id an text support? I feel we should predict what are the customer's requirements and do it in once. Support token id, text and both is one solution for all the customers.
> 
> This already supports token id, text and both as response. User will not input both text and tokens at the same time in 1 request right? Why will user input both text and tokens to request JetStream API?
> 
> I was thinking simplified it to 2 mode:
> 
> 1. For MLPerf, return token ids only
> 2. Other cases, always return text + token ids + score (to be added to Sample)

Looks good to me for these setting:

Request: token id or text (one of)
Response: token id, text or both of them.

FanhaiLu1 (approved):

FanhaiLu1 (commented):
rollback approve

FanhaiLu1 (approved):

FanhaiLu1:
> * When input text, return both text and token ids.

Is it still a streaming mode?

JoeZijunZhou:
> > * When input text, return both text and token ids.
> 
> Is it still a streaming mode?

Yes, each streaming content contains a token_id and its text piece with greedy decode strategy.

PR #82, opened 2024-05-13
Decode Batch Percentage Metrics/Improved Scraping

Bslabe123:
Changes:
- Added metric for available slots in decode batch percentage
- Changed scraping technique for prefill backlog queue metric
- Prometheus client endpoint enablement/port not set via ENV anymore

Results from various prometheus client queries:
```
# HELP jetstream_prefill_backlog_size Size of prefill queue
# TYPE jetstream_prefill_backlog_size gauge
jetstream_prefill_backlog_size{id="maxengine-server-58f8786f4c-4587n"} 57.0
# HELP jetstream_slots_available_percentage The percentage of available slots in decode batch
# TYPE jetstream_slots_available_percentage gauge
jetstream_slots_available_percentage{id="maxengine-server-58f8786f4c-4587n",idx="0"} 0.78125
```
```
# HELP jetstream_prefill_backlog_size Size of prefill queue
# TYPE jetstream_prefill_backlog_size gauge
jetstream_prefill_backlog_size{id="maxengine-server-58f8786f4c-4587n"} 28.0
# HELP jetstream_slots_available_percentage The percentage of available slots in decode batch
# TYPE jetstream_slots_available_percentage gauge
jetstream_slots_available_percentage{id="maxengine-server-58f8786f4c-4587n",idx="0"} 0.375
```
```
# HELP jetstream_prefill_backlog_size Size of prefill queue
# TYPE jetstream_prefill_backlog_size gauge
jetstream_prefill_backlog_size{id="maxengine-server-58f8786f4c-4587n"} 0.0
# HELP jetstream_slots_available_percentage The percentage of available slots in decode batch
# TYPE jetstream_slots_available_percentage gauge
jetstream_slots_available_percentage{id="maxengine-server-58f8786f4c-4587n",idx="0"} 0.96875
```


jetstream/metrics/__init__.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2024 Google LLC
```
JoeZijunZhou, line 1: Could you move this to `jetstream/core/metrics.py` and put the metrics into a class with methods?

jetstream/metrics/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+import prometheus_client
+import shortuuid
+
+instance_uuid = shortuuid.uuid()
+
+# Metrics we care to observe
+prefill_backlog = prometheus_client.Gauge(
+    name="jetstream_prefill_backlog_size",
+    documentation="Size of prefill queue",
+    labelnames=["uuid"],
```
JoeZijunZhou, line 26: Seems it's better if you could get the pod id or container id in GKE environment, so that you could map the metric to the server container or pod. I feel the existing GKE projects may have some references?

jetstream/core/orchestrator.py
```diff
@@ -248,10 +244,9 @@ def __init__(
     # Stage 1
     # At first, a request is placed here in order to get prefilled.
     self._prefill_backlog = queue.Queue()
-    self._prefill_backlog_size_metric = prometheus_client.Gauge(
-        f"jetstream_prefill_backlog_size_{shortuuid.uuid()}",
-        "Size of prefill queue",
-    )
+    jetstream_metrics.prefill_backlog.labels(
```
JoeZijunZhou, line 247: Could you add flag guard to the metric collection for all places? The flag should be consistent with the one in `server_lib.py`.

jetstream/metrics/__init__.py
Bslabe123: We could set it to something like `os.environ['HOSTNAME']`, are we certain that this is always defined? If not should we keep as is or fallback to another value?

jetstream/metrics/__init__.py
JoeZijunZhou: You could init the object in `server_lib.py` or in the driver init, and call it's methods to collect different metrics? 1 server only has 1 driver with 1 jetstream_metrics obj?

FanhaiLu1: +1 for moving code to metrcis.py insteand of __init__.py.  You can use Singleton pattern if only need init once (https://en.wikipedia.org/wiki/Singleton_pattern)

jetstream/core/server_lib.py
```diff
@@ -122,13 +118,28 @@ def run(
   interleaved_mode = (
       len(config.prefill_slices) + len(config.generate_slices) == 0
   )
+
+  # Setup Prometheus server
+  metrics_collector: JetstreamMetricsCollector = None
+  if metrics_server_config is not None and metrics_server_config.port != 0:
```
JoeZijunZhou, line 124: nit: `if metrics_server_config and metrics_server_config.port:`

jetstream/core/orchestrator.py
```diff
@@ -243,15 +243,16 @@ def __init__(
...
 
     # Stages 1-4 represent the life cycle of a request.
     # Stage 1
     # At first, a request is placed here in order to get prefilled.
     self._prefill_backlog = queue.Queue()
-    self._prefill_backlog_size_metric = prometheus_client.Gauge(
-        f"jetstream_prefill_backlog_size_{shortuuid.uuid()}",
-        "Size of prefill queue",
-    )
+    if self._metrics_collector is not None:
```
JoeZijunZhou, line 252: `if self._metrics_collector:`

jetstream/core/config_lib.py
```diff
@@ -52,6 +53,11 @@ class InstantiatedEngines:
   interleaved_engines: List[engine_api.Engine]
 
 
+@dataclasses.dataclass
+class MetricsServerConfig:
```
JoeZijunZhou, line 57: So now we pass in MetricsServerConfig to enable prometheus metrics, could you add a README in `/docs` to describe how to enable and use the prometheus metrics for future users? It would be very helpful for our customers.

JoeZijunZhou (commented):
Could you add unit tests in `/tests` to cover your code changes? Thanks!

FanhaiLu1 (commented):
new version of code looks good to me

FanhaiLu1 (approved):

jetstream/tests/core/test_server.py
```diff
@@ -60,22 +61,26 @@ async def test_server(
     """Sets up a server and requests token responses."""
     ######################### Server side ######################################
     port = portpicker.pick_unused_port()
+    metrics_port = portpicker.pick_unused_port()
```
JoeZijunZhou, line 64: Could you create a separate test to test prometheus metric enabled? 1 test for not enabled, 1 test for enabled.

requirements.txt
```diff
@@ -4,13 +4,14 @@
 #
 #    pip-compile requirements.in
 #
-absl-py==1.4.0
+absl-py==2.1.0
```
JoeZijunZhou, line 7: why do we update the requirements here?

Bslabe123: These changes were the result of running `pip-compile requirements.in` since this is a generated file

docs/online-inference-with-maxtext-engine.md
```diff
@@ -109,6 +109,7 @@ export ICI_TENSOR_PARALLELISM=1
 export SCAN_LAYERS=false
 export WEIGHT_DTYPE=bfloat16
 export PER_DEVICE_BATCH_SIZE=11
+export PROMETHEUS_PORT=9090
```
JoeZijunZhou, line 112: Can we add prometheus metric enabled as a separate example instead of changing the existing config?

requirements.txt
JoeZijunZhou: You are not adding new dep in `requirements.in`, plz don't update the txt file. And plz run with python3.10.

JoeZijunZhou (approved):

PR #84, opened 2024-05-17
Allow tokenizer to customize stop_tokens

qihqi:


jetstream/engine/token_utils.py
```diff
@@ -349,6 +349,11 @@ def bos_id(self) -> int:
     """ID of the BOS token."""
     return self.vocab.bos_id
 
+  @property
```
FanhaiLu1, line 352: Unless we want to return different stop_tokens, we don't need to this function in child class. 

Remove the stop_tokens function here does not affect behavior. By default, it will call parent  stop_tokens.

jetstream/engine/token_utils.py
```diff
@@ -395,6 +400,11 @@ def decode(self, token_ids: list[int]) -> str:
     """
     return self.tokenizer.decode(token_ids)
 
+  @property
```
FanhaiLu1, line 403: same, remove this function in child class will do same thing.

jetstream/engine/token_utils.py
```diff
@@ -349,6 +349,11 @@ def bos_id(self) -> int:
     """ID of the BOS token."""
     return self.vocab.bos_id
 
+  @property
+  def stop_tokens(self) -> set[int]:
+    """ID of the stop token."""
+    return {self.eos_id, self.pad_id}
+
```
FanhaiLu1, line 357: @JoeZijunZhou  What is TikToken? (some tokenizer from tiktok)

jetstream/engine/token_utils.py
qihqi: we want to return different stop_tokens for TikToken: https://github.com/google/JetStream/blob/main/jetstream/third_party/llama3/llama3_tokenizer.py#L94

jetstream/engine/token_utils.py
qihqi: it's llama3's tokenizer (apperently made by OpenAI)

jetstream/engine/token_utils.py
qihqi: done.

bhavya01 (approved):

FanhaiLu1 (approved):

PR #85, opened 2024-05-20
Add profiling support and update docs

JoeZijunZhou:
- Add JAX profiling support based on #16 
- Add docs for profiling instruction and update the observability doc
- Update server unit test for profiling and observability

### DIFF
--- a/jetstream/core/server_lib.py
+++ b/jetstream/core/server_lib.py
@@ -95,6 +95,7 @@ def run(
     threads: int | None = None,
     jax_padding: bool = True,
     metrics_server_config: config_lib.MetricsServerConfig | None = None,
+    enable_jax_profiler: bool = False,
 ) -> JetStreamServer:
   """Runs a server with a specified config.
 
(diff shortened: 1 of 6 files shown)


jetstream/core/server_lib.py
```diff
@@ -148,6 +152,13 @@ def run(
   logging.info("Starting server on port %d with %d threads", port, threads)
 
   jetstream_server.start()
+
+  # Setup Jax Profiler
+  if enable_jax_profiler:
+    logging.info("Starting JAX profiler server on port 9999")
+    jax.profiler.start_server(9999)
```
FanhaiLu1, line 159: Instead of hardcode port number, can you define a profile port flag and use it?

JoeZijunZhou: https://jax.readthedocs.io/en/latest/profiling.html#manual-capture-via-tensorboard, I think the tensorboard is connecting to port 9999 to the jax profiler?

JoeZijunZhou: Updated.

PR #87, opened 2024-05-23
Add ray disaggregated serving support

FanhaiLu1:
This PR add ray disaggregated serving support in JetStream, the underline engine implementation will be pytorch and maxtext side.

This PR doesn't not impact any current interleave or Pathway disaggregated behavior.

### DIFF
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -513,21 +513,19 @@ class Driver:
   def _jax_transfer_prefill_result(self, new_request, target_idx):
     new_request.prefill_result = jax.device_put(
         new_request.prefill_result,
-        self._generate_engines[
-            target_idx
-        ].get_prefix_destination_sharding(),
+        self._generate_engines[target_idx].get_prefix_destination_sharding(),
     )
     # Block here so we don't block on the generate thread that steps.
     jax.block_until_ready(new_request.prefill_result)
- 
+
   def _ray_transfer_prefill_result(self, new_request, target_idx):
     self._generate_engines[target_idx].transfer(new_request.prefill_result)
- 
+
   def _transfer_prefill_result(self, new_request, target_idx):
     if self._ray_multiple_host:
       self._ray_transfer_prefill_result(new_request, target_idx)
-    else:  
-      self._jax_transfer_prefill_result(new_request, target_idx)  
+    else:
+      self._jax_transfer_prefill_result(new_request, target_idx)
 
   def _transfer_thread(self, idx: int):
     """Transfers the kv cache on an active request to the least full
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -223,6 +223,7 @@ class Driver:
       interleaved_mode: bool = False,
       jax_padding: bool = True,
       metrics_collector: JetstreamMetricsCollector | None = None,
+      ray_multiple_host: bool = False,
   ):
     if prefill_engines is None:
       prefill_engines = []
@@ -374,6 +375,7 @@ class Driver:
         )
     )
     self.live = True
+    self._ray_multiple_host = ray_multiple_host
     # Start all threads
     for t in self._all_threads:
       t.start()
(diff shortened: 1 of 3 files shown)


jetstream/core/orchestrator.py
```diff
@@ -223,6 +223,7 @@ def __init__(
       interleaved_mode: bool = False,
       jax_padding: bool = True,
       metrics_collector: JetstreamMetricsCollector | None = None,
+      ray_multiple_host: bool = False,
```
allenwang28, line 226: nit: a better name for this could be with_ray_backend?

jetstream/core/orchestrator.py
```diff
@@ -508,6 +510,23 @@ def _prefill_thread(self, idx: int):
       del prefill_result
       del request
 
+  def _jax_transfer_prefill_result(self, new_request, target_idx):
```
allenwang28, line 513: Could we add the typing annotation for the input and outputs?

allenwang28 (commented):
Looks good, thanks Fanhai! Just minor nits

jetstream/core/server_lib.py
```diff
@@ -97,6 +97,7 @@ def run(
     metrics_server_config: config_lib.MetricsServerConfig | None = None,
     enable_jax_profiler: bool = False,
     jax_profiler_port: int = 9999,
+    ray_multiple_host: bool = False,
```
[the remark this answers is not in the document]
vipannalla: +1

jetstream/core/orchestrator.py
```diff
@@ -508,6 +510,23 @@ def _prefill_thread(self, idx: int):
...
+    new_request.prefill_result = jax.device_put(
+        new_request.prefill_result,
+        self._generate_engines[target_idx].get_prefix_destination_sharding(),
+    )
+    # Block here so we don't block on the generate thread that steps.
+    jax.block_until_ready(new_request.prefill_result)
+
+  def _ray_transfer_prefill_result(self, new_request, target_idx):
+    self._generate_engines[target_idx].transfer(new_request.prefill_result)
```
vipannalla, line 522: I don't see anything ray specfic here, the transfer code is abstracted into engine.transfer(). Can we use a generic name such as "non_jax_transfer" instead of "ray_transfer" here? 
Jetstream doesn't need to know if its Ray or some other mechanism used. Also, lets move this setting to server config for better control.

vipannalla (commented):
Unitests please.

jetstream/core/orchestrator.py
FanhaiLu1: Sounds good and thanks the suggestion! With_ray_backend is better name, then I feel is_ray_backend (we usually use 'is' or 'enable' for bool type parameter prefix). 

Changed to is_ray_backend, how about this name?

jetstream/core/orchestrator.py
FanhaiLu1: Good point, added.

jetstream/core/server_lib.py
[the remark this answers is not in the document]
FanhaiLu1: Thanks! done.

jetstream/core/orchestrator.py
allenwang28: sg! Thanks

jetstream/core/orchestrator.py
FanhaiLu1: Done for the server config part. 

The code here is interface level, the actually logic (or real implementation) is in engine side (pytorch or maxtext). The implementation logic would be: 1: gather prefill result from TPU chips in ray worker 2: transfer all gathered result from TPU to  CPU ram though PCIE in ray worker3: transfer prefill result from prefill server to decode server though DCN by ray head. 

Right now or in near future, I don't see we will introduce another transfer mechanism beside ray or jax, I feel name it as ray_transffer is clear and straightforward.

FanhaiLu1:
> LGTM! May need some unit tests.

Let's explore how to add disaggregated serving unit test and add disaggregated tests before enable disaggregated serving.

JoeZijunZhou:
> > LGTM! May need some unit tests.
> 
> Let's explore how to add disaggregated serving unit test and add disaggregated tests before enable disaggregated serving.

We could add a simple unit test in `test_orchestrator.py` and `test_server.py` just to cover the flag if else logic. The e2e functional test could be added later I guess?

jetstream/core/orchestrator.py
allenwang28: Based on the implementation of `_ray_transfer_prefill_result`, could we assign the responsibility of implementing transfer to the generate engine?

What I mean by this is that within the orchestrator, we can generically use:

```
def _transfer_prefill_result(
      self, new_request: ActiveRequest, target_idx: int
  ):
   self._generate_engines[target_idx].transfer(new_request.prefill_result)
```

and so your Pathways engine needs to implement transfer (the `jax.device_put` logic) and your Ray engine needs to implement transfer with ray.remote primitives.

Or does this break some assumption? I think having JAX involved in the orchestrator is a somewhat leaky abstraction

FanhaiLu1: Agree. @JoeZijunZhou May have more insights, but based on my understanding, this could break Pathway.

FanhaiLu1: Synced with @vipannalla offline, we should removed jax dependencies in JetStream in long term. Current concerns is what is the impact to Pathway, we will figure out and make sure Patyway can adopt no jax in jetstream.

PR #89, opened 2024-05-24
Add JetStream E2E test CI

JoeZijunZhou:
- Added CI to trigger JetStream E2E test for each main branch commit (each PR merged to main branch)
   - https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/290 
- Use a GCP Service Account to execute the command in CI.
   - Added Service Account Key `secrets.GOOGLE_SA_KEY` in the current Github repo
   - https://github.com/google-github-actions/auth?tab=readme-ov-file#sake

Manual trigger E2E test result: https://46c56b977f2b44158661f1ef150119c7-dot-us-central1.composer.googleusercontent.com/dags/jetstream_e2e_inference/grid
- Test Duration: 18 mins

### DIFF
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -11,6 +11,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Details for GCP Service Account Key setup: https://github.com/google-github-actions/auth?tab=readme-ov-file#sake
+
 name: E2E Tests
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e_tests.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+name: E2E Tests
+
+on:
+  push:
+    branches: [main]
+
+permissions:
+  id-token: write
+  contents: read
```
yeandy, line 24: For my education, can you explain why we need permissions here? Is it a requirement for `google-github-actions`?

yeandy (approved):
LGTM

.github/workflows/e2e_tests.yaml
JoeZijunZhou: Yes, following https://github.com/google-github-actions/auth?tab=readme-ov-file#usage.

PR #92, opened 2024-05-29
Model warmup support with AOT and endpoint for JetStream

vivianrwu:
This PR aims to support model warmup on the JetStream server using [Ahead-of-time compilation (AOT)](https://jax.readthedocs.io/en/latest/aot.html). This is to eliminate compile times after the model server has been loaded, allowing the server to serve without extra latencies.

This covers using AOT on prefill, generate, and insert, and using their respective compiles in the prefill and generate threads when serving. A driver flag `self.warmup_enabled` is added to monitor this.

To enable modelwarmup, one can set `--enable_model_warmup=True/true` when running the Maxengine or jetstream-pytorch server. This will enable a wrapper engine over the original engine that will call the respective AOT-compiled prefills, inserts, and generates when appropriate in `prefill_threads` and `generate_threads`

Test coverage can be found under `test_server.py` to check if the model warmup has been successful. 
This functionality will be useful for model server readiness check and pod startup on GKE, to denote that a pod is ready to serve without extra latencies from compilation time.

This has been validated with `Maxtext` on HEAD (https://github.com/google/maxtext/commit/f8ae413536909f07ed3820701380fb2be3169542) on GKE
```
  - model_name=gemma-7b
  - tokenizer_path=assets/tokenizer.gemma
  - per_device_batch_size=1
  - max_prefill_predict_length=1024
  - max_target_length=2048
  - async_checkpointing=false
  - ici_fsdp_parallelism=1
  - ici_autoregressive_parallelism=-1
  - ici_tensor_parallelism=1
  - scan_layers=false
  - weight_dtype=bfloat16
  - load_parameters_path=<ckpt_path>
  - enable_model_warmup=true

curl --request POST --header "Content-type: application/json" -s localhost:8000/generate --data '{
    "max_tokens": 200
}'
{
    "response": " for data science in 2023?\n\n1. Python\n2. R\n3. SQL\n4. Java\n5. Scala\n\n**Note:** The order is based on popularity and demand in the data science industry in 2023."
}
```

The below covers initial testing on jetstream-pytorch and the latency differences

**Latency difference on first request after model server has loaded:**
```
- --size=7b
- --model_name=llama-2
- --batch_size=4
- --max_cache_length=2048
- --quantize_weights=False
- --quantize_kv_cache=False
- tpu platform: v5e-8
```
**Before AOT model warmup (includes compilation): 34.43s**
**After AOT model warmup (no further compilation needed): 1.81s**

This is an initial implementation of AOT for model warmup. With higher batch sizes in jetstream-pytorch, we observe the detokenizing generate step and time to first response after AOT to be slower. **Latency below is measured from the time the request is sent to the time that a response is outputted.**

| Batch size | Time (s) no AOT, before compilation | Time (s) no AOT, after compilation | Time (s) AOT, after compilation | Time (ms) no AOT, detokenizing generate step | Time (ms) AOT, detokenizing generate step |
| ---------- | ----------- | ------------ | ------------ | -------------- | ------------- |
|4 | 34.43 | 1.53 | 1.81 | 5.11 | 6.33 |
| 8 | 36.18 | 1.76 | 3.06 | 6.27 | 9.10 |
| 16 | 34.85 | 2.19 | 4.66 | 7.84 | 14.10 |
| 32 | 36.52 | 2.83 | 5.58 | 11.20 | 24.11 |
| 56 | 39.48 | 3.69 | 9.69 | 15.47 | 42.83 |
| 64 | not recorded | not recorded | OOM | not recorded | OOM |


jetstream/core/orchestrator.py
```diff
@@ -492,12 +495,26 @@ def _prefill_thread(self, idx: int):
...
+      request.true_length = true_length
+
       # Compute new kv cache for the prefill_content.
-      prefill_result = prefill_engine.prefill(
-          params=prefill_params,
-          padded_tokens=padded_tokens,
-          true_length=true_length,
-      )
+      if self.warmup_enabled:
+        padded_token_length = token_utils.take_nearest_length(
```
FanhaiLu1, line 502: In general, I feel the warmup code should be outside of orchstrator. We would like keep orchstrator only contain necessary functions (benchmark or warmup is not necessary function),  make sure the code is clean  and clear. The logic is already very complex to read right now. 

Can you do refactor and move warmup out of this class?

FanhaiLu1:
Thanks for adding the warmup support! Which vm did you run the test?

vivianrwu:
> Thanks for adding the warmup support! Which vm did you run the test?

@FanhaiLu1 I tested using v5e-8, I'll add this detail to the PR comment too.

jetstream/core/orchestrator.py
liurupeng: just curious, do you mean we move the warmup logic to a separate function and then invoke that function here or we call AOT at a completely different place outside of orchestrator?

jetstream/core/orchestrator.py
```diff
@@ -899,3 +944,30 @@ async def HealthCheck(  # pylint: disable=invalid-overridden-method
...
+      request: jetstream_pb2.ModelWarmupRequest,
+      context: Optional[grpc.aio.ServicerContext] = None,
+  ) -> jetstream_pb2.ModelWarmupResponse:
+    """ModelWarmup."""
+    if context is None:
+      logging.warning(
+          "LLM orchestrator is being used in offline test mode, and will not"
+          " respond to gRPC queries - only direct function calls."
+      )
+    # If model warmup wants to be disabled, and we will disable it.
```
liurupeng, line 959: maybe rephrase the comment

jetstream/core/proto/jetstream.proto
```diff
@@ -23,6 +23,8 @@ service Orchestrator {
   rpc Decode(DecodeRequest) returns (stream DecodeResponse) {}
   // Checks if the model server is live.
   rpc HealthCheck(HealthCheckRequest) returns (HealthCheckResponse) {}
+  // Warms up the model server.
+  rpc ModelWarmup(ModelWarmupRequest) returns (ModelWarmupResponse) {}
```
liurupeng, line 27: @FanhaiLu1 I was wondering would we need to expose a new API for switching warmup, in the long term, I was assuming that this will be enabled by default, and adding a new API field may be hard to deprecate later. But there is no other obvious downside as well

jetstream/core/orchestrator.py
vivianrwu: I moved the warmup code out of orchestrator but kept the check (if self.warmup_enabled) + its corresponding logic because of the following functionality: Once model warmup is called, its compiled form (prefill, insert, and generate), will be stored in their respective dictionaries for their corresponding bucket length. This compiled form should be called from now on, or else the JetStream server will experience compilation times.

jetstream/core/proto/jetstream.proto
vivianrwu: I think given that there is the performance diff at the moment, we would need to expose a new API for enabling and disabling warmup. Would deprecation process be difficult in the future? I would assume that at the moment we wouldn't enable using model warmup by default. @FanhaiLu1

jetstream/core/orchestrator.py
FanhaiLu1: Correct me if I'm wrong, we know what type of data prefill or decode should be process for warmup. In this case, all the code can be outside orchestrator. 

@JoeZijunZhou Please also take a look. The orchestrator is already complex, we's better to keep this class only have main function code, other wise, it's hard to maintain and refactor in future.

jetstream/core/proto/jetstream.proto
FanhaiLu1: It's depends. even in long term, this is option user can chose do warm up or not.

jetstream/core/orchestrator.py
JoeZijunZhou: +1. I feel it's feasible to implement a wrapper for the engines, and pass the compiled engines etc in the driver init here: https://github.com/google/JetStream/blob/main/jetstream/core/server_lib.py#L141. Then, we don't need to change the orchestrator and the engine API, making the AOT warmup logics decoupled from the existing jetstream core and engine.

JoeZijunZhou (commented):
Great work! I feel we could refactor the warmup logic to make it clean and decoupled.

jetstream/core/orchestrator.py
vivianrwu: After the refactor, the AOT warmup backend logic is mostly decoupled from orchestrator. If warmup is invoked, after that, all prefills will have to use the AOT compiled prefill, or else JetStream will recompile, and this endpoint supports enable / disable model warmup. So this means, when a user send a decode request after warming up, it *has* to use the compiled prefill instead of the normal prefill function. If a user decides they don't want to use the model warmup feature anymore and they disable, then when decoding it will use the normal prefill function. 

Is it possible to pass compiled engines in the driver after the driver is already initialized? Since warmup occurs after the JetStream server has completed loading. 

If there is a way during Decode to check if model server warmup has been enabled, and then to use the compiled prefill, insert, and decode, outside of orchestrator, then that would be great! But I agree, if possible, it would be great to decouple this from orchestrator for easy maintenance.

jetstream/engine/token_utils.py
```diff
@@ -33,6 +33,21 @@
...
 ResultTokens = Any
 
+DEFAULT_PREFILL_BUCKETS = [
+    16,
+    32,
+    64,
+    128,
+    256,
+    512,
+    1024,
```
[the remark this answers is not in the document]
vivianrwu: It wouldn't compile sequence length 32768 because of this logic: https://github.com/google/JetStream/pull/92/files/6702d2f136226b1b0417ac7cf8412540d7ef0f29#diff-38fa7646a82d9a00519f201908d44aca60d5db5db38613e73d2b668ec01bb239R72-R75, it identifies the max_prefill_length and sets prefill_buckets accordingly.

jetstream/core/orchestrator.py
JoeZijunZhou: I guess you could add the if else switching engine logics (what you added in the orchestrator) in the wrapper methods? In you wrapper object, you could store the original engines and compiled engines. WDYT?

qihqi: You can probably put the logic of warm up into another `Engine` implementation that takes an instance of Engine instance:

```python

class WarmedUpEngine(engine_api.Engine):

   def __init__(self, downstream_engine: engine_api.Engine):
       # do compile, setup the dicts that maps int to jax Compiled.
  
   def prefill(self, ...):
       return self.compiled_prefill[seqlen](*args, **kwargs) etc
  # same for insert / generate

```


Then, 

in orchestrator, you only need

```python

if warmed up:
   self.engine = WarmedUpEngine(self.engine)
```
in __init__ and rest dont need to change.

jetstream/core/proto/jetstream.proto
qihqi: Let's just put this into `service Orchestrator`

jetstream/engine/aot_utils.py
```diff
@@ -0,0 +1,238 @@
...
+      if bucket <= prefill_engine.max_prefill_length
+  ]
+  prefill_engine.prefill_buckets = prefill_buckets
+  if prefill_engine.max_prefill_length not in prefill_buckets:
+    prefill_buckets.append(prefill_engine.max_prefill_length)
+
+  def compile_prefill(length):
+    metadata = prefill_engine.get_tokenizer()
+    vocab = token_utils.load_vocab(metadata.path, metadata.extra_ids)
+    padded_tokens, true_length = token_utils.tokenize_and_pad(
```
qihqi, line 84: The shape of padded tokens should be of [batch size x seq len]. So you can do `jnp.ones((batch size, length), dtype='int32')` (or maybe int64 please verify first) instead

jetstream/engine/aot_utils.py
```diff
@@ -0,0 +1,238 @@
...
+
+  generate_compiled = compile_generate()
+  logging.info(
+      "---------Generate engine %d compiled generation step.---------",
+      generate_idx,
+  )
+  generate_engine.generate_compiled = generate_compiled
+
+  insert_compiled = {}
+  with concurrent.futures.ThreadPoolExecutor(
```
qihqi, line 227: try run without this to see if it still OOMs?

jetstream/engine/engine_api.py
```diff
@@ -134,6 +134,14 @@ class Engine(abc.ABC):
   JetStream efficient serving infrastructure.
   """
 
+  # Compiled prefills
+  prefill_compiled: dict[int, jax.stages.Compiled]
```
qihqi, line 138: Shouldnt be part of interface. Please remove

jetstream/core/orchestrator.py
JoeZijunZhou: Yep the wrapper object is a wrapper of Engine (implementation of Engine API), just like adding a layer on top of the original engine with the warmup handling. You could also store the warmup state in the wrapper, then you don't need to change orchestrator at all I guess.

vivianrwu: Ah, I see, this makes sense! Thank you! I will try this

jetstream/core/proto/jetstream.proto
vivianrwu: Should we put this in `service Orchestrator` or its own service? I thought we want to decouple model warmup from orchestrator entirely? @FanhaiLu1

jetstream/engine/aot_utils.py
```diff
@@ -0,0 +1,238 @@
...
+  prefill_buckets = [
+      bucket
+      for bucket in prefill_buckets
+      if bucket <= generate_engine.max_prefill_length
+  ]
+  generate_engine.prefill_buckets = prefill_buckets
+  if generate_engine.max_prefill_length not in prefill_buckets:
+    prefill_buckets.append(generate_engine.max_prefill_length)
+
+  decode_state = generate_engine.init_decode_state()
```
qihqi, line 152: this line will allocate empty kv caches
Then engine will allocate kv caches again.

To compile, you don't need the real input. You can give model jax.ShapedDtypeArray instead. Only shape and dtype need to be specified.

jetstream/core/proto/jetstream.proto
FanhaiLu1: Its own service is better.

vivianrwu: Thinking about this again, do you mean instead of having an endpoint, bake the AOT warmup logic into init based on if a user passes in a flag on model server startup say `--warmup = True`? This would remove the need of a model server warmup endpoint, but if a user decides they want to enable warmup, then they would have to restart deployment of the model server.

FanhaiLu1: --warmup = True with a flag looks good. Engineer who start the server know they want to warmup or not warmup.

vivianrwu: Sure, I think once we fix the OOM / performance gap we can switch to use flag.

jetstream/core/orchestrator.py
vivianrwu: Added two things: 
1. Logic to bake the warmup into model server startup
2. Wrapper engine definition `WarmedUpEngine`

Added some extra logic to help facilitate the engine / define the warm up state, since that is used later on in prefill threads and generate threads to determine which bucket is needed to called with.

jetstream/engine/aot_utils.py
vivianrwu: Yes, int32 is correct. However when using batch_size in the jnp.ones, there is an error with mismatch shape later on. Using without batch size seems to work. Does using batch size perhaps help with the performance degradation issues? Since using without batch size works for now, closing and we can address this in a future PR

```
File "/jetstream-pytorch/jetstream_pt/layers.py", line 365, in reshape_for_broadcast
    assert freqs_cis.shape == (
AssertionError: freqs_cis: (1, 256, 4), x: (1, 256, 32, 64)
```

jetstream/core/orchestrator.py
vivianrwu: @FanhaiLu1 Added the wrapper logic, ptal and let me know if it looks good! Thanks. We can have a follow up PR to address the performance degradation.

JoeZijunZhou: I feel it's feasible to move the warmup state and its related handling into `WarmedUpEngine`, WDYT?

jetstream/engine/aot_utils.py
```diff
@@ -0,0 +1,246 @@
...
+      "---------Generate engine %d compiled generation step.---------",
+      generate_idx,
+  )
+  generate_engine.generate_compiled = generate_compiled
+
+  insert_compiled = {}
+  with concurrent.futures.ThreadPoolExecutor(
+      max_workers=len(prefill_buckets)
+  ) as executor:
+    _ = list(executor.map(compile_insert, prefill_buckets))
```
qihqi, line 238: insert compiled is not updated here?

jetstream/engine/aot_utils.py
```diff
@@ -0,0 +1,246 @@
...
+
+    prefill, _ = prefill_engine._downstream_engine.prefill(  # pylint: disable=protected-access
+        params=prefill_params,
+        padded_tokens=padded_tokens,
+        true_length=true_length,
+    )
+
+    lowered = jax.jit(generate_engine._downstream_engine.insert).lower(  # pylint: disable=protected-access
+        prefix=prefill, decode_state=decode_state, slot=1
+    )
```
qihqi, line 180: can you try printing out lowered.as_text() here? So we want to see if slot is hardcoded in the generated StableHLO or is captured as variable. If it's hardcoded then this line is going to recompile

jetstream/core/orchestrator.py
FanhaiLu1: +1 for Zijun's comments

vivianrwu: Thanks, decoupled the warmup handling per our discussion offline!

jetstream/engine/aot_utils.py
vivianrwu: It is updated in the compile_insert, but I changed it to be updated outside now!

jetstream/engine/aot_utils.py
vivianrwu: Discussed offline, slot is captured as variable

JoeZijunZhou (approved):
Synced with @vivianrwu , will do complete refactor in orchestrator.py and server_lib.py in the following PRs. Approve for the init PR.

FanhaiLu1 (commented):
Thanks for the changes!

PR #94, opened 2024-05-30
Enable JetStream Standalone Server

JoeZijunZhou:


FanhaiLu1 (commented):
what is the purpose of adding a standalone server?

JoeZijunZhou:
> what is the purpose of adding a standalone server?

To improve the usability of JetSteam Server. Eventually we could have a single command to run the server with different engines, different models, different configs.

PR #96, opened 2024-06-03
Update benchmark config for xlml automation

morgandu:
- Enabled optional full warmup 
- Updated benchmark serving flags type for flag sweeping through xlml

### DIFF
--- a/benchmarks/benchmark_serving.py
+++ b/benchmarks/benchmark_serving.py
@@ -81,6 +81,31 @@ import pandas
 from eval_accuracy import eval_accuracy
 
 
+def str2bool(v: str) -> bool:
+  """Convert a string of truth to True or False.
+
+  Args:
+    - v (str):
+      - True values are 'y', 'yes', 't', 'true', and '1';
+      - False values are 'n', 'no', 'f', 'false', and '0'.
+
+  Returns:
+    bool: True or False
+
+  Raises:
+    ValueError if v is anything else.
+  """
+  v = v.lower()
+  true_values = ["y", "yes", "t", "true", "1"]
+  false_values = ["n", "no", "f", "false", "0"]
+  if v in true_values:
+    return True
+  elif v in false_values:
+    return False
+  else:
+    raise ValueError(f"Invalid value '{v}'!")
+
+
 @dataclass
 class BenchmarkMetrics:
   """Data class to store benchmark metrics."""
@@ -226,9 +251,9 @@ def tokenize_dataset(
 
 def filter_dataset(
     tokenized_dataset: list[tuple[str, Any, str, int, int, int]],
-    max_output_length: Optional[int] = None,
+    max_output_length: int = 0,
 ) -> list[InputRequest]:
-  if max_output_length is None:
+  if max_output_length != 0:
     print("In InputRequest, pass in actual output_length for each sample")
   else:
     print(
@@ -269,7 +294,7 @@ def sample_requests(
     dataset: list[tuple[Any, Any]],
     tokenizer: Any,
     num_requests: int,
-    max_output_length: Optional[int] = None,
+    max_output_length: int = 0,
     oversample_multiplier: float = 1.2,
 ) -> list[InputRequest]:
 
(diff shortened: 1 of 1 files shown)


benchmarks/benchmark_serving.py
```diff
@@ -576,7 +601,10 @@ def main(args: argparse.Namespace):
 
   if args.warmup_first:
     print("Warm up start:")
-    warmup_requests = list(sample_warmup_requests(input_requests)) * 2
+    if args.full_warmup:
```
JoeZijunZhou, line 604: What is the full_warmup here? Why does the `warmup_requests = list(sample_warmup_requests(input_requests)) * 2` not work?

benchmarks/benchmark_serving.py
```diff
@@ -81,6 +81,31 @@
 from eval_accuracy import eval_accuracy
 
 
+def str2bool(v: str) -> bool:
```
JoeZijunZhou, line 84: I remember @yeandy had made the type conversion in xlml pipeline. Do we need this in benchmark script?

benchmarks/benchmark_serving.py
```diff
@@ -711,9 +737,9 @@ def main(args: argparse.Namespace):
   parser.add_argument(
       "--request-rate",
       type=float,
-      default=float("inf"),
+      default=0.0,
```
JoeZijunZhou, line 740: Why do we change it to 0.0?

benchmarks/benchmark_serving.py
yeandy: I don't think we made a type conversion. `benchmark_serving.py` probably interprets this as a string "true" rather than a `True` boolean.

https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/inference/configs/maxtext_inference_gce_config.py#L132

JoeZijunZhou: I see, does the original implementation works in the xlml? If yes, do we still need this conversion?

morgandu: We will still need this conversion.
The problem is when you pass in --warmup-first false, which is interpreted as True

The only way of not warmup is not use this flag. That's why the conversion helps here when a boolean flag need to sweep across True and False.

Warmup is just one of the examples here.

benchmarks/benchmark_serving.py
morgandu: Similar to the boolean flag, if I am sweeping through different request-rate, I can not set float("inf"), because it is being recognized as string when passing in to benchmark_serving.py.

benchmarks/benchmark_serving.py
morgandu: Full warmup gets better perf result than the existing warmup, shared with you the detail results offline

yeandy (approved):
LGTM

benchmarks/benchmark_serving.py
```diff
@@ -792,16 +819,22 @@ def main(args: argparse.Namespace):
...
+      type=str2bool,
       default=False,
       help="Whether to run evaluation script on the saved outputs",
   )
   parser.add_argument(
       "--warmup-first",
-      type=bool,
+      type=str2bool,
       default=False,
       help="Whether to send warmup req first",
```
vipannalla, line 830: We should combine with "--warmup-first" option and instead add new option  "--warmup-mode" with possible values = [None, sampled, full]. We already have so many options, need to stop the bloat :).

morgandu: Done

benchmarks/benchmark_serving.py
```diff
@@ -599,12 +599,14 @@ def main(args: argparse.Namespace):
...
     )
 
-  if args.warmup_first:
-    print("Warm up start:")
-    if args.full_warmup:
-      warmup_requests = input_requests
-    else:
-      warmup_requests = list(sample_warmup_requests(input_requests)) * 2
+  warmup_requests = None
+  if args.warmup_mode == "full":
```
JoeZijunZhou, line 603: Should we make it to `num_decode_slots * (len([16, 32, 64, 128, 256, 512, 1024]) = batch_size * num_chips * 7` to ensure the server is completely warmup?

morgandu: Let's follow up with diff PR for correct/fixing different warmup mode

PR #93, opened 2024-05-30
chore: update model_ckpt_conversion.sh

eltociear:
Covert -> Convert

PR #97, opened 2024-06-03
Add tensorboard plugin dep for remote access

JoeZijunZhou:
- Add `tensorboard-plugin-profile` dep to enable tensorboard remote access
- Include typo fix #93 Thanks @eltociear


FanhaiLu1 (approved):
Thanks for adding it!

PR #105, opened 2024-06-25
Prefill return first token

jwyang-google:
Modify Jetstream to make prefill return first token. Pending testing with MLPerf loadgen.

### DIFF
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -512,12 +512,11 @@ class Driver:
       request.prefill_result = prefill_result
 
       # put first token to detokenize queue
-      request.complete = np.zeros(
-        (prefill_engine.samples_per_slot,), np.bool_
-      )
+      request.complete = np.zeros((prefill_engine.samples_per_slot,), np.bool_)
       my_detokenize_backlog = self._detokenize_backlogs[idx]
       my_detokenize_backlog.put(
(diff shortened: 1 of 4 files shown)


jetstream/core/orchestrator.py
```diff
@@ -714,7 +724,34 @@ def _detokenize_thread(self, idx: int):
       if data is None:
         break
       start_detokenize_time = time.time()
-      if isinstance(data[1], engine_api.ResultTokens):
+      # prefill first token
+      if isinstance(data[0], engine_api.ResultTokens):
```
JoeZijunZhou, line 728: Could we merge the prefill first token block into the generate step tokens block? Since the detokenization logic is identical.

JoeZijunZhou (commented):
Need to update unit tests and resolve the CICD failures after all.

jetstream/core/orchestrator.py
jwyang-google: The logic are actually different. The generate step are processing all the slots and the available slots are depending on which slots are put into the detokenize_backlog. If we do the same thing for prefill, both prefill and generate threads are putting slot information and tokens into detokenize log, this will make the logic much more complicated. I think leave prefill as separate block making the code easier to understand.

JoeZijunZhou (approved):
LGTM! Thanks Junwei!

PR #112, opened 2024-07-15
Cleanup orchestrator proto

JoeZijunZhou:
- Removed unused fields in orchestrator proto
- Update all files according to the proto cleanup

### DIFF
--- a/jetstream/core/proto/jetstream.proto
+++ b/jetstream/core/proto/jetstream.proto
@@ -26,9 +26,6 @@ service Orchestrator {
 }
 
 message DecodeRequest {
-  // Where to load any pre-existing kv cache from.
-  string session_cache = 1;
-  int32 priority = 3;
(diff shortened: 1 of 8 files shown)


jetstream/core/proto/jetstream.proto
```diff
@@ -26,9 +26,6 @@ service Orchestrator {
 }
 
 message DecodeRequest {
-  // Where to load any pre-existing kv cache from.
-  string session_cache = 1;
```
FanhaiLu1, line 30: In future, we will have prefill cache logic. Is this filed related with prefill cache?

JoeZijunZhou: This is for history cache originally. We could add prefill cache fields when we implement it.

FanhaiLu1: Sounds good!

FanhaiLu1 (approved):
LGTM

PR #115, opened 2024-07-18
Add http server to JetStream

JoeZijunZhou:
- Add Http server entrypoint to JetStream
- default: `http://localhost:8080/`
- API docs: `http://localhost:8080/docs`
- generate API: `POST http://localhost:8080/v1/generate, DecodeRequest JSON`
- health check API: `GET http://localhost:8080/v1/health`
- Tested with mock engine
- TODO: Integrate with MaxText and PyTorch engines after #94

### DIFF
--- a/jetstream/entrypoints/http/api_server.py
+++ b/jetstream/entrypoints/http/api_server.py
@@ -12,17 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""JetStream Http API server."""
+
 import json
 from typing import Sequence
-from absl import app
+from absl import app as abslapp
 from absl import flags
 from fastapi import APIRouter, Response
 import fastapi
-from fastapi.responses import JSONResponse, StreamingResponse
+from fastapi.responses import StreamingResponse
 import uvicorn
+from google.protobuf.json_format import Parse
 
 from jetstream.core import config_lib, orchestrator, server_lib
+from jetstream.core.proto import jetstream_pb2
 from jetstream.entrypoints.config import get_server_config
+from jetstream.entrypoints.http.protocol import DecodeRequest
+from jetstream.entrypoints.http.utils import proto_to_json_generator
 
 flags.DEFINE_string("host", "0.0.0.0", "server host address")
 flags.DEFINE_integer("port", 8080, "http server port")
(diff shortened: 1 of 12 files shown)


jetstream/entrypoints/http/api_server.py
```diff
@@ -0,0 +1,118 @@
...
+@router.get("/")
+def root():
+  """Root path for Jetstream HTTP Server."""
+  return Response(
+      content=json.dumps({"message": "JetStream HTTP Server"}, indent=4),
+      media_type="application/json",
+  )
+
+
+@router.post("/v1/generate")
```
FanhaiLu1, line 61: Right now, we don't have multiple version of generate. Can we replace request url "/v1/generate" with "generate"? 

Even there is multiple version generation, I suggest not use v1, v2 to expose the url, chose a meaningful name in this case.

jetstream/entrypoints/http/api_server.py
```diff
@@ -0,0 +1,118 @@
...
+
+def server(argv: Sequence[str]):
+  # Init Fast API.
+  app = fastapi.FastAPI()
+  app.include_router(router)
+
+  # Init LLMOrchestrator which would be the main handler in the api endpoints.
+  devices = server_lib.get_devices()
+  print(f"devices: {devices}")
+  server_config = get_server_config(flags.FLAGS.config)
```
FanhaiLu1, line 91: The server_config from get_server_config are both test config. I'm wondering we should expose a method to let engine side pass the engine and devices.

JoeZijunZhou: It will be TODO: Integrate with MaxText and PyTorch engines after https://github.com/google/JetStream/pull/94. If we add them here now, it would fail due to the engine deps are not added to JetStream yet.

jetstream/entrypoints/http/api_server.py
JoeZijunZhou: I am referring to the SOTA LLM APIs for the API versioning:
- [Gemini API](https://ai.google.dev/gemini-api/docs/api-versions?_gl=1*yv9jv8*_up*MQ..&gclid=CjwKCAjw1920BhA3EiwAJT3lSbkW3p63bNIcQCVr7kPEaAGdATpOTg2gvSUpoZTDJTprBpu2AHsV5xoCH54QAvD_BwE)
- [Openai API](https://platform.openai.com/docs/api-reference/chat/create)
- [vllm API](https://docs.vllm.ai/en/latest/getting_started/quickstart.html#using-openai-completions-api-with-vllm)

We could remove the version now, but eventually adding the version should be the best practice IIUC.

jetstream/entrypoints/http/api_server.py
FanhaiLu1: Sounds good, thanks!

FanhaiLu1 (approved):

PR #117, opened 2024-07-19
Fix benchmark script for saving benchmark result

lsy323:
if `args.additional_metadata_metrics_to_save` is None, `json.loads` would fail. `args.additional_metadata_metrics_to_save` is None by default. Adding a check to avoid the failure.

Test:
Using the benchmark example script in `README`
```
python JetStream/benchmarks/benchmark_serving.py   \
--tokenizer ~/maxtext/assets/tokenizer.llama2  \
--warmup-mode full   \
--save-result   \
--save-request-outputs   \
--request-outputs-file-path outputs.json   \
--num-prompts 1000   \
--max-output-length 1024   \
--dataset openorca
```


FanhaiLu1 (approved):
Thanks for fixing it!

JoeZijunZhou (approved):

PR #118, opened 2024-07-19
Add `jetstream_server_startup_latency` metric

Bslabe123:
Validated in Cloud Monitoring: Screenshot 2024-07-22 at 10 09 06 AM

### DIFF
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -384,9 +384,9 @@ class Driver:
             self.detokenize_threads,
         )
     )
-    if self.metrics_collector:
-      self.metrics_collector.get_server_startup_latency_metric().set(
-          time.time() - metrics_collector.server_start_time
+    if self._metrics_collector:
+      self._metrics_collector.get_server_startup_latency_metric().set(
+          time.time() - self._metrics_collector.server_start_time
       )
     self.live = True
     self._is_ray_backend = is_ray_backend
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -384,10 +384,6 @@ class Driver:
             self.detokenize_threads,
         )
     )
-    if self._metrics_collector:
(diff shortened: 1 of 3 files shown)


jetstream/core/server_lib.py
```diff
@@ -145,6 +137,21 @@ def run(
         "Not starting Prometheus server: --prometheus_port flag not set"
     )
 
+  logging.info("Kicking off gRPC server.")
```
FanhaiLu1, line 140: Looks like you only need the "get_model_load_time_metric()" change, any reason to move all gRPC server code?

Bslabe123: Actually just moved the prometheus server startup code to before the grpc server code since the metric tracks the time it takes for part of it to run

jetstream/core/orchestrator.py
```diff
@@ -384,6 +384,10 @@ def __init__(
             self.detokenize_threads,
         )
     )
+    if self.metrics_collector:
```
JoeZijunZhou, line 387: Are you collecting the model load time? Why adding the metric here?

Bslabe123: Synced with @liurupeng, for our purposes we want to measure the time between server start and when the server is live (when self.live = True).

jetstream/core/server_lib.py
FanhaiLu1: sounds good, thanks.

jetstream/core/orchestrator.py
```diff
@@ -384,6 +384,10 @@ def __init__(
             self.detokenize_threads,
         )
     )
+    if self._metrics_collector:
+      self._metrics_collector.get_server_startup_latency_metric().set(
+          time.time() - self._metrics_collector.server_start_time
+      )
```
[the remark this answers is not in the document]
Bslabe123: Actually yeah that would make things simpler, let me add that!

JoeZijunZhou (approved):

PR #122, opened 2024-07-26
Performance optimized interleaved mode JetStream server

JoeZijunZhou:
- Optimized TPU duty cycle (largest gap < 4ms)
- Optimized TTFT: dispatch prefill tasks ASAP w/o unnecessary blocking in CPU, keep backpressure to enforce insert ASAP, return first token ASAP.
- Optimized TPOT: properly enforce generate and detokenize task in sequential w/o unnecessary blocking in CPU.
- Optimized output token throughput: properly prioritize prefill and balancing TTFT and decode in high throughput situation.

- Tested with llama2-70b JetStream MaxText server on v5e-8 VM

### DIFF
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -201,6 +201,7 @@ class Driver:
   # This can be a list because we can pass it as an arg to generate and
   # detokenize threads. It is a list of tokens to be detokenized.
   _detokenize_backlogs: list[queue.Queue[engine_api.ResultTokens]] = []
+  _prefill_detokenize_backlogs: list[queue.Queue[engine_api.ResultTokens]] = []
   _generate_slots: list[queue.Queue[int]] = []
   _active_requests: list[queue.Queue[tuple[int, ActiveRequest]]] = []
 
@@ -316,6 +317,12 @@ class Driver:
         queue.Queue(8)
         for _ in self._generate_engines
     ]
+    self._prefill_detokenize_backlogs = [
+        # We don't let detokenization accumulate more than 8 steps to avoid
+        # synchronization issues.
+        queue.Queue(8)
+        for _ in self._prefill_engines
+    ]
 
     # A queue of integers representing available 'slots' in the decode
     # operation. I.e. potentially available rows in the batch and/or microbatch.
@@ -376,12 +383,23 @@ class Driver:
         )
         for idx in range(len(self._generate_engines))
     ]
+    self.prefill_detokenize_threads = [
+        JetThread(
+            target=functools.partial(
+                self._prefill_detokenize_thread,
+                idx,
+            ),
+            name=f"prefill-detokenize-{idx}",
+        )
+        for idx in range(len(self._prefill_engines))
+    ]
     self._all_threads = list(
         itertools.chain(
             self._prefill_threads,
             self._transfer_threads,
             self._generate_threads,
             self.detokenize_threads,
+            self.prefill_detokenize_threads,
         )
     )
     self.live = True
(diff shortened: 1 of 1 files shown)


jetstream/core/orchestrator.py
```diff
@@ -316,6 +317,12 @@ def __init__(
         queue.Queue(8)
         for _ in self._generate_engines
     ]
+    self._prefill_detokenize_backlogs = [
+        # We don't let detokenization accumulate more than 8 steps to avoid
```
FanhaiLu1, line 321: Can you elaborate more on why there is synchronization issue after 8 steps?

jetstream/core/orchestrator.py
```diff
@@ -795,6 +792,44 @@ def _detokenize_thread(self, idx: int):
         slot, active_request = data
         my_live_requests[slot] = active_request
 
+  def _prefill_detokenize_thread(self, idx: int):
```
FanhaiLu1, line 795: I though we already had prefill detokenize thread. Do current jetstream (before this pr) always return prefill token (fist token) after first decode step?

FanhaiLu1:
> * Optimized TPU duty cycle (largest gap < 4ms)
> * Optimized TTFT: dispatch prefill tasks ASAP w/o unnecessary blocking in CPU, keep backpressure to enforce insert ASAP, return first token ASAP.
> * Optimized TPOT: properly enforce generate and detokenize task in sequential w/o unnecessary blocking in CPU.
> * Optimized output token throughput: properly prioritize prefill and balancing TTFT and decode in high throughput situation.
> * Tested with llama2-70b JetStream MaxText server on v5e-8 VM

Optimized TTFT and Optimized output token throughput are conflicted with each. Can we expose some parameter to tuning the two part?

jetstream/core/orchestrator.py
JoeZijunZhou: We only had detokenize thread that combined prefill detokenize and decode detokenize. The problem is that we have `jax.block_until_ready()` blocking the thread waiting for the prefill token or decode token copy to host async, so putting them in 1 thread would make the TTFT slow. JetStream returns prefill token in prefill thread (after prefill step generating the first token).

jetstream/core/orchestrator.py
JoeZijunZhou: Set it to 8 as the detokenize thread. Too large or too small will cause performance issue.

JoeZijunZhou:
> > * Optimized TPU duty cycle (largest gap < 4ms)
> > * Optimized TTFT: dispatch prefill tasks ASAP w/o unnecessary blocking in CPU, keep backpressure to enforce insert ASAP, return first token ASAP.
> > * Optimized TPOT: properly enforce generate and detokenize task in sequential w/o unnecessary blocking in CPU.
> > * Optimized output token throughput: properly prioritize prefill and balancing TTFT and decode in high throughput situation.
> > * Tested with llama2-70b JetStream MaxText server on v5e-8 VM
> 
> Optimized TTFT and Optimized output token throughput are conflicted with each. Can we expose some parameter to tuning the two part?

Currently, prioritize prefills in interleaved mode, and apply correct JAX blocking for copy to host async to reduce wasted wait time. 1 more optimization to do is to ensure the result returns immediately when the return channel has the result (from orchestrator).

jetstream/core/orchestrator.py
FanhaiLu1: sounds good, thanks for sharing insights!

jetstream/core/orchestrator.py
FanhaiLu1: Thanks, is this PR ready to submit?

PR #125, opened 2024-08-06
Makefile

Bslabe123:
Added a makefile to cover a few common use cases:
- `update-deps`: generate the requirements.txt and install dependencies from it
- `generate-protos:`: update all protos the repo
- `check`: run the `pylint` linter
- `format`: format code


Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+generate-protos: generate-and-prepend-preambles format
+
+format:
+	$(PIP) install pyink
+	pyink --pyink-indentation 2 --line-length 80 --verbose .
+
+check:
+	pylint --ignore-patterns=".*_pb2.py,.*_pb2_grpc.py" jetstream/ benchmarks/
+
```
FanhaiLu1, line 21: Thanks for adding this one! What is the benefit of it compared with current approach? It seems more complex.

.github/workflows/unit_tests.yaml
```diff
@@ -53,7 +53,7 @@ jobs:
         pytype --jobs auto --disable=import-error,module-attr jetstream/ benchmarks/
     - name: Analysing the code with pylint
       run: |
-        pylint jetstream/ benchmarks/
+        make check
```
JoeZijunZhou, line 56: Awesome! Do you mind also replacing other commands in this CI? Thanks!

Bslabe123: Sure!

Makefile
Bslabe123: For adding the --ignore-patterns or for adding the target entirely? Added the --ignore-patterns flag so we dont need to add `pylint: disable=all` to files when we regenerate protos. Added the target entirely to make is usable elsewhere, for example if we want to add a precommit hook

.github/workflows/unit_tests.yaml
Bslabe123: Done

JoeZijunZhou (approved):
Have we formatted the Makefile?

Bslabe123:
> Have we formatted the Makefile?

No standard way to format a makefile as far as I can see

PR #126, opened 2024-08-07
Manual model warmup to resolve AOT model warmup performance degradation

vivianrwu:
Use manual model warmup instead of AOT implemented model warmup, since with AOT, we observe performance degradation at higher batch size of maxtext configuration, mentioned in https://github.com/google/JetStream/pull/92:
1. OOM at higher batch size (after model warmup, during an active request)
2. Slower detokenizing generate step time exponentially at higher batch sizes

This has been verified that the detokenizing generate step time remains same as JetStream optimal behavior for all batch sizes. 

```
curl --request POST --header "Content-type: application/json"
 -s localhost:8000/generate --data '{
    "max_tokens": 200
}'
{
    "response": " for data science in 2023?\n\n1. Python\n2. R\n3. SQL\n4. Java\n5. Scala\n\n**Note:** The order is based on popularity and demand in the data science industry in 2023."
}
```

### DIFF
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -134,7 +134,6 @@ class ActiveRequest:
   prefill_result: Any = None
   #################### Information relevant for prefill ########################
   prefill_content: Optional[str | list[int]] = None
-  padded_token_length: Optional[int] = None
   ################## Information relevant for detokenization ###################
   # Which generate step this was added at.
   generate_timestep_added: Optional[int] = None
@@ -509,7 +508,6 @@ class Driver:
           padded_tokens=padded_tokens,
           true_length=true_length,
       )
-
       request.prefill_result = prefill_result
 
       # put first token to detokenize queue
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -502,11 +502,6 @@ class Driver:
       padded_tokens, true_length = self._process_prefill_content(
           request, tokenizer, is_bos, prefill_engine.max_prefill_length
       )
-      if isinstance(prefill_engine, engine_api.JetStreamEngine):
-        request.padded_token_length = token_utils.take_nearest_length(
-            prefill_engine.prefill_buckets, true_length
-        )
-        prefill_engine.set_padded_token_length(request.padded_token_length)
 
       # Compute new kv cache for the prefill_content.
       prefill_result, first_token = prefill_engine.prefill(
@@ -677,11 +672,6 @@ class Driver:
             slot,
             generate_timestep,
         )
-
-        if isinstance(generate_engine, engine_api.JetStreamEngine):
-          generate_engine.set_padded_token_length(
-              new_request.padded_token_length
-          )
 
         decode_state = generate_engine.insert(
             new_request.prefill_result, decode_state, slot=slot
--- a/jetstream/engine/warmup_utils.py
+++ b/jetstream/engine/warmup_utils.py
@@ -66,7 +66,7 @@ def layout_params_and_compile_executables(
     )
     inserts_generate_compiled.append([insert_generate_compiled])
 
-  if any(prefills_compiled) and any(inserts_generate_compiled):
+  if all(prefills_compiled) and all(inserts_generate_compiled):
     return True
   return False
 
--- a/jetstream/engine/warmup_utils.py
+++ b/jetstream/engine/warmup_utils.py
@@ -66,7 +66,7 @@ def layout_params_and_compile_executables(
     )
     inserts_generate_compiled.append([insert_generate_compiled])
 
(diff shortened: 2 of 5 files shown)


jetstream/core/orchestrator.py
```diff
@@ -502,11 +502,6 @@ def _prefill_thread(self, idx: int):
       padded_tokens, true_length = self._process_prefill_content(
           request, tokenizer, is_bos, prefill_engine.max_prefill_length
       )
-      if isinstance(prefill_engine, engine_api.JetStreamEngine):
-        request.padded_token_length = token_utils.take_nearest_length(
```
JoeZijunZhou, line 506: can we remove padded_token_length: https://github.com/google/JetStream/pull/92/files#diff-753df909960695fd6aa61b95c4762c4edc50a298007d40de7096a7e49e0e1215R138, and the extra new line: https://github.com/google/JetStream/pull/92/files#diff-753df909960695fd6aa61b95c4762c4edc50a298007d40de7096a7e49e0e1215R519

JoeZijunZhou (commented):
Do we need to update unit tests? 

QQ on the description, 
1. we set the max pdbs when we start the server, this value should be within memory cap (based on calculation w the devices used), then it would not OOM right?
2. Why higher actual batch size would have very slow detokenization? Could you share some investigation or profiles?

vivianrwu:
> Do we need to update unit tests?
Unit tests do not need to be updated because it is on the condition of engine.warm
> QQ on the description,
> 
> 1. we set the max pdbs when we start the server, this value should be within memory cap (based on calculation w the devices used), then it would not OOM right?
Yes, I think the storage of the compiled graphs from AOT and executing it from AOT is what takes up the memory. We observe the OOM at generate request.
> 2. Why higher actual batch size would have very slow detokenization? Could you share some investigation or profiles?
Yes, you can reference https://github.com/google/JetStream/pull/92 for some investigations. Also shared the doc internally.

jetstream/engine/warmup_utils.py
```diff
@@ -44,34 +43,30 @@ def layout_params_and_compile_executables(
...
+        prefill_engine=any_prefill_engine,
+        generate_engine=ge,
+        prefill_params=any_prefill_params,
+        generate_params=generate_params[i],
+        generate_idx=i,
     )
+    inserts_generate_compiled.append([insert_generate_compiled])
 
-  if prefill_executables and inserts_generate_executables:
+  if prefills_compiled and inserts_generate_compiled:
```
JoeZijunZhou, line 69: should be `if any(prefills_compiled) and any(inserts_generate_compiled):` since they are lists?

vivianrwu: Thanks, changed!

FanhaiLu1:
> ified that the detokenizing generate step time remains same as JetStream optimal behavior for all batch sizes.

Did you figure out what is the root cause of  performance issue and OOM for AOT?

jetstream/core/orchestrator.py
```diff
@@ -678,11 +671,6 @@ def _generate_thread(self, idx: int):
             generate_timestep,
         )
 
-        if isinstance(generate_engine, engine_api.JetStreamEngine):
```
FanhaiLu1, line 681: I assume there is not warmup code in orchestrator, correct me if I'm wrong.

vivianrwu: There is none, since there is not a need for it. Since we are not saving any compiled graphs (like how we did with AOT).

vivianrwu:
> > ified that the detokenizing generate step time remains same as JetStream optimal behavior for all batch sizes.
> 
> Did you figure out what is the root cause of performance issue and OOM for AOT?

RCA has been attempted and the root cause of OOM can potentially be the added space to save the compiled graphs in executables alongside saving the cache in the compilation cache directory. The performance issue, has not been concluded. Could be unoptimal AOT executables. I can share the investigation offline

jetstream/engine/warmup_utils.py
```diff
@@ -44,34 +43,30 @@ def layout_params_and_compile_executables(
...
+        prefill_engine=any_prefill_engine,
+        generate_engine=ge,
+        prefill_params=any_prefill_params,
+        generate_params=generate_params[i],
+        generate_idx=i,
     )
+    inserts_generate_compiled.append([insert_generate_compiled])
 
-  if prefill_executables and inserts_generate_executables:
+  if any(prefills_compiled) and any(inserts_generate_compiled):
```
JoeZijunZhou, line 69: Sorry, I guess I mistake my previous comment, the condition should be all truth for both prefills_compiled and inserts_generate_compiled IIUC? Then we should use all() instead of any()

vivianrwu: Thanks, changed!

JoeZijunZhou (approved):

PR #127, opened 2024-08-07
Refactor Prometheus Metrics Logic and Added to Docs

Bslabe123:


jetstream/core/orchestrator.py
```diff
@@ -545,17 +538,10 @@ def _prefill_thread(self, idx: int):
...
-      if self._metrics_collector:
-        self._metrics_collector.get_time_per_prefill_token().observe(
-            (
-                request.metadata.transfer_enqueue_time
-                - request.metadata.prefill_dequeue_time
-            )
-            / true_length
-        )
+      get_metric("jetstream_request_input_length").observe(true_length)
+      get_metric("jetstream_time_per_prefill_token").observe(
```
JoeZijunZhou, line 542: Can we wrap it into 1 function like `collect_jetstream_time_per_prefill_token_metric`? Also apply it to all other metrics.

PR #129, opened 2024-08-09
Standalone JetStream removes pinned deps

JoeZijunZhou:


FanhaiLu1 (approved):
Looks good, thanks for making the change!

PR #132, opened 2024-08-19
Update JetStream instructions

yeandy:
- General cleanup of instructions
- Generalizing the script to convert llama checkpoints s.t. it supports custom GCS buckets for the maxtext checkpoints.
- Adding quantization instructions


FanhaiLu1 (approved):
Thanks for the changes!

PR #136, opened 2024-09-05
Change the default message for requester.py and remove mlperf 4.1 install for proxy version support.

zhihaoshan-google:
using a more positive default message for requester.py.


vipannalla (approved):
Thanks

PR #152, opened 2024-11-18
add seperate prefill detokenization thread

zhihaoshan-google:
Add a seperate prefill detokenization thread to make sure the detokenization of prefill and decode is not blocked with each other.

### DIFF
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -38,7 +38,7 @@
     of the generation loop at the relevant slot.
    - Regardless, it performs a step.
   - It takes the sampled tokens, and places them on a 'detokenizing_queue'.
-7. Within the detokenizing thread:
+7. Within the detokenizing thread (Prefill and Generate separately):
   - Tokens are detokenized for every 'slot' in a given set of sampled tokens.
   - When an end condition is met, the 'slot' integer is returned to the
     respective generation queue.
@@ -210,7 +210,8 @@ class Driver:
   # Stage 4
   # This can be a list because we can pass it as an arg to generate and
   # detokenize threads. It is a list of tokens to be detokenized.
-  _detokenize_backlogs: list[queue.Queue[engine_api.ResultTokens]] = []
+  _prefill_detokenize_backlogs: list[queue.Queue[engine_api.ResultTokens]] = []
+  _generate_detokenize_backlogs: list[queue.Queue[engine_api.ResultTokens]] = []
   _generate_slots: list[queue.Queue[int]] = []
   _active_requests: list[queue.Queue[tuple[int, ActiveRequest]]] = []
 
@@ -270,11 +271,11 @@ class Driver:
     # one of the generate backlogs.
     # Interleaved Mode: Max size is 1 to increase the HBM utilization
     # during generate.
-    # Disaggregated Mode: Max size is 4 to allow for 2 prefills to be enqueued
-    # while 1 transfer is enqueued while 1 is being transferred.
+    # Disaggregated Mode: Max size is 16 to allow for total 16 prefills to
+    # be enqueued or enqueued while 1 is being transferred.
     # TODO: Make queue size configurable.
     self._transfer_backlogs = [
-        queue.Queue(1 if self._interleaved_mode else 4)
+        queue.Queue(1 if self._interleaved_mode else 16)
         for i in range(len(self._prefill_engines))
     ]
     if self._metrics_collector:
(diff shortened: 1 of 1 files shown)


jetstream/core/orchestrator.py
```diff
@@ -376,13 +387,25 @@ def __init__(
...
+            ),
+            name=f"prefill_detokenize-{idx}",
+        )
+        for idx in range(len(self._generate_engines))
+    ]
+    self.generate_detokenize_threads = [
         JetThread(
             target=functools.partial(
                 self._detokenize_thread,
+                False,
```
patemotter, line 405: Nit: add keyword `is_prefill` when creating the threads since it's not obvious what True/False correspond to at a glance.

patemotter (commented):
Overall LGTM, just some clarification question/suggestion.

jetstream/core/orchestrator.py
```diff
@@ -270,11 +271,11 @@ def __init__(
     # one of the generate backlogs.
     # Interleaved Mode: Max size is 1 to increase the HBM utilization
     # during generate.
-    # Disaggregated Mode: Max size is 4 to allow for 2 prefills to be enqueued
-    # while 1 transfer is enqueued while 1 is being transferred.
+    # Disaggregated Mode: Max size is 16 to allow for 15 prefills to be enqueued
+    # while 1 is being transferred.
```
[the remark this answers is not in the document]
zhihaoshan-google: Rephased it as "Max size is 16 to allow for total 16 prefills to be enqueued or enqueued while 1 is being transferred"

jetstream/core/orchestrator.py
zhihaoshan-google: Thanks for the advice! Done.

patemotter (approved):

zhihaoshan-google:
Thank you, Pate!

patemotter:
`jetstream/core/orchestrator.py:274:74: C0303: Trailing whitespace (trailing-whitespace)`

:-/

zhihaoshan-google:
I will make sure test pass before submission. Sorry for the mistake.

vipannalla:
Looks good. What improvement do you see with this change?

vipannalla (approved):

zhihaoshan-google:
> Looks good. What improvement do you see with this change?

it's in the analysis doc.

wangkuiyi:
Hi Zhihao and Pate,

I think this PR may disrupt the correct detokenization order. Please let me know if my understanding is incorrect.

The proposed change creates separate detokenization backlogs: P backlogs for prefill engines and G backlogs for generation engines. For each active request, the first token (from the prefill thread) and subsequent tokens (from the generate thread) end up in different backlogs. This separation could lead to incorrect ordering.

To maintain proper token ordering, all tokens from a single request should be processed through the same backlog before being returned via the request's channel.

Consider this edge case: if Python's thread scheduler never executes the detokenization threads that read from the prefill-detokenization backlogs, only the tokens from generate threads would be returned. This would result in missing first tokens for all requests.

zhihaoshan-google:
Thank you for the comment, Yi!

Yes, you are right. Sorry for missing the potential breaks.  

The previous problem before this PR from my understanding is:
[
Long time to convert from jax Array to numpy array
](url)



I am currently thinking of two options:

1. Still use two threads: add a condition variable per request to make sure the order (waiting for the first token to add in the request return channel before adding the second and following token).  

2. Change back to one thread: Wait for prefill and generation/decode jitted function to complete inside the prefill and generation/decode thread directly instead of in the detokenization thread. Also to mitigate the runtime latency (for dispatching the device program), we can schedule several decode/generate steps together. (this might be most common way for this issue)

I might need to do a quick evaluation and get back to you.

wangkuiyi:
How about merging the P prefill detokenization backlogs and the G generation detokenization backlogs into a single set of N detokenization backlogs? These backlogs will be associated with N detokenization threads?

For any active request r, any prefill or generation thread processing r simply adds the output token IDs to the i-th of these N tokenization backlogs, where i is calculated as the hash of r modulo N.

This ensures that all response tokens for r go to the same detokenization backlog.  This also ensures that all tokens for r arrive in the correct order.

wangkuiyi:
Thank you, Zhihao, for the context and the quick reaction! Looking forward to your evaluation result!

changlan:
Hi Zhihao and Vipan, just following up on this issue - do we have a conclusion for this, or has the change been reverted already? It is blocking our update due to potential incorrect results. Thanks.

vipannalla:
@sixiang-google, can you take a look?

PR #161, opened 2025-01-07
An experimental JAX inference framework for prototyping new ideas.

zhihaoshan-google:
An experimental JAX inference framework for prototyping new ideas. 

So far it uses the llama2-7b chat model as the proof of concept. It should be helpful for the further JAX inference framework development. It is currently implemented without changing or using any JetStream API so that the further refactoring and integration is needed. It can also provide the GPU support by replacing several TPU kernels (KV Cache write, prefill and decode attention and collective matmul) with the GPU version (either CUDA or Triton implementation).

It has the following features so far (some of them are limited version):

Performance:
1. Paged Attetnion
2. Chunked Prefill and Piggybacking Decode
3. Bidirectional Collective Matmul

Framework:
1. Pythonic model builder (Consider migrating to FLAX NNX when the https://github.com/google/flax/issues/4261 get fixed instead of using the current implementation)
2. JAX manual sharding
3. Interface for different hardware support
4. On-the-flying HF model conversion and deployment

This offline number is around `45%` better than the current MaxText and JetStream (as of 2024/08/16) number in the same situation (llama2-7b chat model with 8-ways TP on v5e-8 without any quantization).

```
Benchmarking result:
  Total requests: 1000
  Total input tokens: 218743
  Total output tokens: 291740
  Input token throughput: 2980.654636529649 tokens/sec
  Output token throughput: 3975.332621666338 tokens/sec
```

Note: The online number should be even more better than the current MaxText and JetStream because the experimental framework runs the prefill and decode together in one model forward pass.

Please refer to README file for more information and how to reproduce (only takes 10 mins and several simple steps).

Will add more reports in the later commits.


zhihaoshan-google:
Given the experimental folder is not related with the JetStream and existing code, I am going to merge it directly.

tohaowu (approved):

tohaowu (approved):

vipannalla (approved):
Thanks, this is all experimental and should be good to merge.

PR #164, opened 2025-01-15
[Minor Bug] Fix a bug in the number of detokenize thread (found by fhzhang)

wyzhang:
Issue:
self.prefill_detokenize_threads should have the same # of threads as # of prefill engines. 
While existing code creates num-of-generate-engines detokenize threads


vipannalla (approved):

vipannalla:
Good catch

PR #172, opened 2025-02-03
Add test for benchmark_serving

deqiangc:
Add test for benchmark_serving.  
Additionally, fix two minor issues:
1. sort import statement of benchmark_serving lexicographically per group
2. Add missing transformer requirments
3. Due to the addition of this new test, some module that are not checked against coverage are now checked and we disable those coverage for now.


vipannalla (approved):
Looks good, thanks for adding these tests

PR #174, opened 2025-02-03
[Minor Bug] Fix a bug in the blocking algorithm of the generate thread.

Lumosis:
`_transfer_backlogs` isn't 1:1 with `generate_threads` so we cannot use `generate_thread`'s index to access _transfer_backlogs. Instead, we should block when any `transfer_backlog` is non-empty.


sixiang-google:
LGTM

vipannalla (approved):

PR #182, opened 2025-02-06
Add comprehensive debugging messages.

Lumosis:
To enable the comprehensive debug logging, run `> LOG_LEVEL={DEBUG|INFO|WARNING}` before starting Jetstream

### DIFF
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -98,20 +98,23 @@ from jetstream.engine import engine_api, tokenizer_api, token_utils
 from jetstream.core.metrics.prometheus import JetstreamMetricsCollector
 import numpy as np
 
-root = logging.getLogger()
-root.setLevel(logging.WARNING)
+log_level = os.getenv("LOG_LEVEL", "WARNING").upper()
+
+logger = logging.getLogger("JetstreamLogger")
+logger.propagate = False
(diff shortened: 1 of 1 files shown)


jetstream/core/orchestrator.py
```diff
@@ -246,19 +251,26 @@ def __init__(
       is_ray_backend: bool = False,
   ):
     if prefill_engines is None:
+      logger.warning("No prefill engines provided.")
```
wyzhang, line 254: Should we throw exception?

wyzhang: Same for below.

jetstream/core/orchestrator.py
```diff
@@ -711,34 +758,41 @@ def _generate_thread(self, idx: int):
...
+              f"Put back the slot.",
+          )
           # If we were blocking and hit the timeout, then retry the loop.
           # Otherwise, we can exit and proceed to generation.
           if block:
             continue
           else:
             break
 
-        # Signal to kill the thread.
```
wyzhang, line 721: Intended to delete if new_request is None return?

Lumosis: Yes, was about to have a separate PR for this one. But yeah, we can delete this in this PR

PR #193, opened 2025-02-18
[Deepseek] enable mmlu benchmark for deepseek

bzgoogle:
Create gen_mmlu_prompt and load_mmlu_dataset_csv for mmlu dataset


benchmarks/deepseek/benchmark_serving.py
```diff
@@ -0,0 +1,1075 @@
+# Copyright 2024 Google LLC
```
vipannalla, line 1: Can you add the mmlu dataset functionality to JetStream/benchmarks/benchmark_serving.py?  Deepseek is the first one, but not the only model to use mmlu dataset and eval.

benchmarks/deepseek/benchmark_serving.py
```diff
@@ -0,0 +1,1075 @@
...
+  outputs = []
+  for _, row in samples.iterrows():
+    prompts.append(row["input"])
+    outputs.append(row["output"])
+
+  return [(prompt, output) for prompt, output in zip(prompts, outputs)]
+
+def gen_mmlu_prompt(
+    method: str, 
+    data: Any,
```
singh-mitali, line 282: Lets just call this row consistently.

benchmarks/deepseek/benchmark_serving.py
```diff
@@ -0,0 +1,1075 @@
...
+    example_data: Any):
```
singh-mitali, line 283: call it example rows.

benchmarks/deepseek/benchmark_serving.py
```diff
@@ -0,0 +1,1075 @@
...
+      question = data[0],
+      choiceA = data[1],
+      choiceB = data[2],
+      choiceC = data[3],
+      choiceD = data[4],
+    )
+  
+  examples = '' 
+  
+  if method == 'HELM':
```
singh-mitali, line 309: Please add unit test case for each of these. Also does the MMLU test use the same examples for all rows? If so we can generate them once instead of each prompt.

benchmarks/benchmark_serving.py
```diff
@@ -334,7 +416,7 @@ def filter_dataset(
       output_len,
       sample_idx,
   ) in tokenized_dataset:
-    if prompt_len < 4 or output_len < 4:
+    if prompt_len < 4 or (not is_mmlu_dataset and output_len < 4):
```
singh-mitali, line 419: Pass these as flags.

benchmarks/benchmark_serving.py
```diff
@@ -276,6 +276,87 @@ def load_openorca_dataset_pkl(
...
+      output += f"({row['answer']}) {content}\n\n"
+
+  return output
+    
+def gen_mmlu_prompt(
+    dataset_path: str,
+    num_shots: int = 1,
+    mmlu_method: str = 'HELM'   
+):
+  combined_dataset, subject_set = load_mmlu_dataset_csv(dataset_path)
```
singh-mitali, line 341: Rename to subject_set to prompts_per_subject

benchmarks/benchmark_serving.py
```diff
@@ -276,6 +276,87 @@ def load_openorca_dataset_pkl(
 
   return [(prompt, output) for prompt, output in zip(prompts, outputs)]
 
+def load_mmlu_dataset_csv(
+    dataset_path: str
+) -> list[tuple[Any, Any]]:
+  if not dataset_path:
+    dataset_path = "./"
```
singh-mitali, line 283: We should assert here. Any default value should be in the flag.

singh-mitali (approved):
Looks much better - Minor nits. Also remove the older redundant files.

singh-mitali (approved):
Please add tests in a subsequent CL.

singh-mitali (approved):

PR #202, opened 2025-02-24
Fix the benchmark_serving minor error.

wang2yn84:


singh-mitali (approved):
Looks good. Please also run the default prior benchmarks before submitting to make sure those still work.

PR #203, opened 2025-02-24
1> Add mmlu dataset; 2> fix minor bug for mmlu prompt; 3> add scikit-…

bzgoogle:
1> Add mmlu test dataset, from  "git clone https://huggingface.co/datasets/Stevross/mmlu"
2> fix minor bug for mmlu prompt; 
3> add scikit-learn to requirements.in


singh-mitali (commented):
Looks good. I would only consider adding a small sampled version of the dataset for testing, else we can skip adding it.

singh-mitali (approved):

PR #205, opened 2025-02-25
Fix the output length issue of Math 500 dataset.

wang2yn84:
Temporarily lifts the output length constraints for Math 500 dataset. Added a TODO for the fix that supports filter based on output size. This temp solutions favors accuracy over performance.

### DIFF
--- a/benchmarks/benchmark_serving.py
+++ b/benchmarks/benchmark_serving.py
@@ -437,6 +437,9 @@ def filter_dataset(
     if prompt_len > 1024 or prompt_len + output_len > 2048:
       # Prune too long sequences.
       continue
+    # TODO: should pass in the actual solution length instead of hardcode.
+    if dataset_type == "math500":
+       output_len = 1024
     request = InputRequest(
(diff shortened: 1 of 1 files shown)


benchmarks/benchmark_serving.py
```diff
@@ -437,6 +437,9 @@ def filter_dataset(
     if prompt_len > 1024 or prompt_len + output_len > 2048:
       # Prune too long sequences.
       continue
+    # TODO: should pass in the actual solution length instead of hardcode.
+    if dataset_type == "math500":
```
singh-mitali, line 441: What is this value based on? Is this the max len for outputs in the dataset ?

singh-mitali: Note some these conditions were based on mlperf datasets - even the condition on line 437-439 above. I don't think we should apply these to the new datasets without understanding impact.

wang2yn84:
Replace this with a more thorough fix PR#207

PR #206, opened 2025-02-25
Fix the eval for Math 500 dataset.

wang2yn84:


singh-mitali (approved):

wang2yn84:
Replace this with a more thorough fix PR#207

PR #209, opened 2025-02-26
Add request id into active request

liurupeng:
Effect of the change:This PR will add an optional args (request_id) in the JetStream's engine API (prefill, insert)
Motivation/Reason: This change will add a unique id for each active request, this would be helpful for tracking the request status in the prefill & insert stage. Per the design to enable PagedAttention with JetStream, this will be needed for requests placement/tracking.

### DIFF
--- a/jetstream/core/orchestrator.py
+++ b/jetstream/core/orchestrator.py
@@ -86,6 +86,7 @@ import sys
 import threading
 import time
 import traceback
+import uuid
 from typing import Any, AsyncIterator, Optional, Tuple, cast, List
 
 import grpc
@@ -147,6 +148,8 @@ class ActiveRequest:
   """Current state of the driver."""
 
   #################### Information relevant for generation #####################
+  # The unique id for the activeRequest, used for tracking the request's status
+  request_id: uuid.UUID
   max_tokens: int
   # We keep prefill and decode information together in the same object so that
   # there is less indirection about where this return channel is.
@@ -564,6 +567,7 @@ class Driver:
           params=prefill_params,
           padded_tokens=padded_tokens,
           true_length=true_length,
+          request_id=request.request_id,
       )
       request.prefill_result = prefill_result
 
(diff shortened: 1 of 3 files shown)


sixiang-google (approved):

jetstream/engine/mock_engine.py
```diff
@@ -101,14 +102,19 @@ def load_params(self) -> Params:
...
+      static_argnames=("request_id",),
+  )
   def prefill(
       self,
       *,
       params: Params,
       existing_prefix: Optional[jax.Array] = None,
       padded_tokens: jax.Array,
       true_length: int,
+      request_id: uuid.UUID = None,
```
wyzhang, line 117: Should be Optional[uuid.UUID] = None, right?

jetstream/engine/engine_api.py
```diff
@@ -181,6 +183,7 @@ def insert(
       prefix: Prefix,
       decode_state: DecodeState,
       slot: int,
+      request_id: uuid.UUID = None,
```
wyzhang, line 186: same here (as mentioned below)

jetstream/engine/engine_api.py
```diff
@@ -143,6 +144,7 @@ def prefill(
       padded_tokens: jax.Array,
       true_length: int,
       sampler: Optional[Callable[[Any], Any]] = None,
+      request_id: uuid.UUID = None,
```
wyzhang, line 147: same here. Optional

jetstream/core/orchestrator.py
```diff
@@ -147,6 +148,8 @@ class ActiveRequest:
   """Current state of the driver."""
 
   #################### Information relevant for generation #####################
+  # The unique id for the activeRequest
```
wyzhang, line 151: Add some comments on how it is used and why it is needed

wyzhang: The code below indicates this can be None. If so, have the right annotation

liurupeng: Sure, will add more comments. The request id itself will not be none, but we could pass none into the engine api

jetstream/engine/engine_api.py
liurupeng: thanks, will be update asap!

wyzhang (approved):

PR #215, opened 2025-02-28
Add multi-sampling functionality.

Lumosis:
With multi-sampling, multiple responses will be generated given one prompt. Clients can specify num_samples in the request.


vipannalla:
Looks ok at a high level, but this change really needs full range of unittests and integration tests.

PR #218, opened 2025-03-05
Kill the generate thread when getting a "None" new request

Lumosis:
If we don't return immediately when getting a None, we may fail the assertion in line 970


vipannalla:
Accepting since this is fixing a bug, but can you add unitests for these cases?

vipannalla (approved):

Lumosis:
> Accepting since this is fixing a bug, but can you add unitests for these cases?

PR #228, opened 2025-03-19
[Performance] Not set request uuid to resolve performance regression

wyzhang:
Why:
When set request uuid, it can trigger jax.jit recompilation for maxtext due to request uuid being unique, resulting in performance regression.


liurupeng:
thank you so much @wyzhang for helping fix this!

liurupeng (approved):

vipannalla (approved):

PR #230, opened 2025-03-21
Support long context dataset accuracy measurement.

Lumosis:
The result should be like:
```

Results

{'rougeL': 6.165264881957181, 'exact_match': 0.0, 'gen_len': 59242, 'gen_num': 50}
```

### DIFF
--- a/benchmarks/benchmark_serving.py
+++ b/benchmarks/benchmark_serving.py
@@ -71,6 +71,7 @@ from typing import Any, AsyncGenerator, Optional
 
 from benchmarks.eval_accuracy import eval_accuracy
 from benchmarks.eval_accuracy_mmlu import eval_accuracy_mmlu
+from benchmarks.eval_accuracy_longcontext import eval_accuracy_longcontext
 from benchmarks.metrics import CounterMetric, EventMetric
 import grpc
 from jetstream.core.proto import jetstream_pb2
@@ -166,6 +167,7 @@ class InputRequest:
   output: str = ""
   output_len: int = 0
   sample_idx: int = -1
+  metric: str = ""
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


vipannalla:
The results you pasted -- are they from an actual benchmark run on 405b? Can you paste the full results (as a screenshot or paste link)?

Lumosis:
> The results you pasted -- are they from an actual benchmark run on 405b? Can you paste the full results (as a screenshot or paste link)?

No, this is from a mock run. I am working on the actual benchmarking.

vipannalla:
sounds good

benchmarks/eval_accuracy_longcontext.py
```diff
@@ -0,0 +1,160 @@
...
+          ]
+      )
+      / len(pred_uuids)
+      * 100
+  )
+
+  return {"exact_match": round(score, 2)}
+
+
+def qa_em(label, pred):
```
mailvijayasingh, line 63: please add some docustring about each functions.

benchmarks/benchmark_serving.py
```diff
@@ -1080,7 +1086,7 @@ def main(args: argparse.Namespace):
           args.dataset_path,
       )
     elif args.dataset == "longcontext":
-      dataset = load_longcontext_dataset_pkl(
+      dataset, metrics = load_longcontext_dataset_pkl(
```
mailvijayasingh, line 1089: can you also add a filter on max supported length? 
We may want to sometimes run on a smaller context. 
So, filter out the prompt if the len(tokens) > max_supported_length?

Lumosis: I believe `max_input_length` can be used to do the filtering. WDYT?

mailvijayasingh: max_input_length - I am not able to recall, does this filter out the prompts or truncate the prompts?

Lumosis: It will filter out the prompts: https://github.com/AI-Hypercomputer/JetStream/blob/main/benchmarks/benchmark_serving.py#L453

vipannalla (approved):
Looks good

PR #233, opened 2025-04-02
Added explicit pathwaysutils.initialize() call to JetStream

wstcliyu:
# Description

pathwaysutils will soon require a call to initialize().

# Tests

Depending on GitHub actions for tests

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [X] I have performed a self-review of my code.
- [X] I have necessary comments in my code, particularly in hard-to-understand areas.
- [X] I have run end-to-end tests tests and provided workload links above if applicable.
- [X] I have made or will make corresponding changes to the doc if needed.

### DIFF
--- a/jetstream/engine/__init__.py
+++ b/jetstream/engine/__init__.py
@@ -15,9 +15,14 @@
 """Initialization for any Engine implementation."""
 
 import jax
+from importlib import util
 
-try:
+if util.find_spec("pathwaysutils"):
   import pathwaysutils
-except ImportError as e:
-  print("Proxy backend support is not added")
-  pass
+
+  pathwaysutils.initialize()
+else:
+  print(
+      "Running JetStream without Pathways. "
+      "Module pathwaysutils is not imported."
+  )
--- a/jetstream/tests/engine/test_init.py
+++ b/jetstream/tests/engine/test_init.py
@@ -1,0 +1,72 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


vivianrwu (commented):
I think this is less of an "error" and more of "we are running jetstream without pathways. Could we clarify too that this could be intended behavior, or else users may think that there is some failure happening.

wstcliyu:
> I think this is less of an "error" and more of "we are running jetstream without pathways. Could we clarify too that this could be intended behavior, or else users may think that there is some failure happening.

Thanks. Updated the log to "Running JetStream without Pathways. Module pathwaysutils is not imported." Please review again.

vipannalla (approved):
looks good, can you fix the lint issue?

vivianrwu (approved):
Thanks for the fixes, lgtm!

wstcliyu:
It's weird that it could not pass the checks (pyink format check & test coverage report). I don't see a quick fix due to lack of change suggestions. Shall we bypass the checks?

vivianrwu:
> It's weird that it could not pass the checks (pyink format check & test coverage report). I don't see a quick fix due to lack of change suggestions. Shall we bypass the checks?

pyink you can fix by running `pyink --pyink-indentation 2 --line-length 80 <file_name>`

wstcliyu:
Thank you. Could you also guide me how to add test coverage for `pathwaysutils.initialize()`. pathwaysutils is not in the requirements file so this line is never called really.

vivianrwu:
> Thank you. Could you also guide me how to add test coverage for `pathwaysutils.initialize()`. pathwaysutils is not in the requirements file so this line is never called really.

You could probably write a test expecting the error.

wstcliyu:
@vivianrwu I added the unit test for module engine init. This unit test can pass by running separately `coverage run -m unittest -v jetstream.tests.engine.test_init` but cannot pass by `coverage run -m unittest -v`

vipannalla:
Its complaining about unit test coverage for a test file, which should really be excluded from the check.

PR #235, opened 2025-04-04
Modify MMLU prompt template to improve accuracy on DeepSeek models; E…

bzgoogle:
Modify MMLU prompt template to improve accuracy on DeepSeek models;
Enable running on whole dataset when setting num_prompts=-1

### DIFF
--- a/benchmarks/benchmark_serving.py
+++ b/benchmarks/benchmark_serving.py
@@ -938,7 +938,7 @@ def parse_args() -> argparse.Namespace:
   parser.add_argument(
       "--num-prompts",
       type=int,
-      default=1000,
+      default=-1,
       help=(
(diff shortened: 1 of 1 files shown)


benchmarks/benchmark_serving.py
```diff
@@ -1133,11 +1133,16 @@ def main(args: argparse.Namespace):
     # A given args.max_output_length value is the max generation step,
     # when the args.max_output_length is default to None, the sample's golden
     # output length will be used to decide the generation step.
+    if args.num_prompts == -1:
```
mitalisi, line 1136: Thanks for making this change. Can we also set default flag value to -1.

mitalisi (approved):
Looks good - one minor change requested.

PR #236, opened 2025-04-07
Remove Jetstream UT check from Add Label

jyj0w0:
Remove Jetstream UT check from Add Label

### DIFF
--- a/.github/workflows/add_label.yaml
+++ b/.github/workflows/add_label.yaml
@@ -1,0 +1,118 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


vipannalla:
Why do you want to bypass the unitest check?

vipannalla:
Do you know which modules are not passing the coverage check? A better option is to reduce the coverage limit temporarily  to 90% instead of disabling it.

vipannalla:
I've update the test coverage to 90% in https://github.com/AI-Hypercomputer/JetStream/pull/237.

jyj0w0:
> I've update the test coverage to 90% in #237.

Thanks! I would close this PR then.

PR #257, opened 2025-04-24
Update ResultToken API to optionally return token log probability 

wenxindongwork:
This PR updates the ResultToken class with an additional `log_prob` attribute which stores token log probabilities. This attribute is set to None by default. 

MaxEngine is updated to populate this attribute when `return_log_prob` is set to `True` in global config. (https://github.com/AI-Hypercomputer/maxtext/pull/1626)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,10 @@ This project follows
 
 ## Contribution process
 
+### Style Formatting
+
+Please run `make format` before submitting your PR. 
+
 ### Code Reviews
 
 All submissions, including submissions by project members, require review. We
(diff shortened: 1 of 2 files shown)


mitalisi (commented):
We dont always have to return logits. Would be better to have a flag for it.

jyj0w0 (approved):

jyj0w0 (approved):

wenxindongwork:
This PR could not get merged. Perhaps because I'm using a fork. Trying this PR with a branch instead (https://github.com/AI-Hypercomputer/JetStream/pull/260)

PR #263, opened 2025-04-29
Add llama benchmarks

mailvijayasingh:


.github/workflows/run_maxtext_jetstream_tests.yaml
```diff
@@ -87,6 +87,21 @@ jobs:
           bash -c "
             bash JetStream/.github/workflows/test_moe_benchmarks.sh
           "
+    - name: Test llama 70b Benchmarks
+      run: |
+        rm -rf ${OUTPUT_DIR}
+        mkdir -p ${OUTPUT_DIR}
```
yuyanpeng-google, line 93: rm -rf ${OUTPUT_DIR}
        mkdir -p ${OUTPUT_DIR}

This two command is setup of the output. We need a separate step at the first, or it will delete the MOE results.

yuyanpeng-google (approved):

.github/workflows/run_maxtext_jetstream_tests.yaml
mailvijayasingh: good catch! Updated that!

jyj0w0 (approved):

PR #268, opened 2025-05-07
feat: Add prefix cache benchmark

yuyanpeng-google:
This commit introduces a new benchmark to test the performance of prefix caching in JetStream.

The benchmark (`benchmark_prefix_cache.sh`) allows testing with various prompt lengths and common prefix lengths. It utilizes a new mock dataset generated by `load_mock_prefix_cache_test_input_requests` in `benchmark_serving.py`, which creates prompts sharing common prefixes of varying lengths based on a normal distribution.

Key changes include:
- New script `benchmarks/benchmark_prefix_cache.sh` to orchestrate prefix cache benchmark runs.
- Added `PrefixCacheTestTokenizer` for simple character-to-ordinal tokenization, suitable for controlled prefix testing.
- Implemented `load_mock_prefix_cache_test_input_requests` in `benchmark_serving.py` to generate test data with shared prefixes.
- Added `prefix_cache_test` as a dataset option and `--prefix-cache-test-common-len` argument to `benchmark_serving.py`.
- Updated `benchmarks/README.md` with instructions on how to run the new prefix cache benchmark.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -75,6 +75,7 @@ python benchmark_serving.py \
 ```
 
 ## Benchmark with openorca dataset (openorca is used by MLPerf inference for LLaMA2 models)
+
 ```
 python JetStream/benchmarks/benchmark_serving.py   \
 --tokenizer ~/maxtext/assets/tokenizer.llama2  \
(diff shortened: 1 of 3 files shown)


vipannalla (approved):
Looks good

PR #269, opened 2025-05-07
Refactor(PrefixCache): New load API, per-layer Tries, async ops & stats

yuyanpeng-google:
Add async to prevent device_get blocking on the critical paths waiting prefill result. Use per-layer tries to prevent load cache from DRAM when common length tie. Add statistic for debug and benchmark.

### DIFF
--- a/jetstream/core/prefix_cache.py
+++ b/jetstream/core/prefix_cache.py
@@ -27,6 +27,8 @@ Key Features:
   the longest matching prefix key based on input tokens.
 - **Thread Safety:** The `PrefixCache` class uses a lock to ensure thread-safe
   operations.
+- **Performance Statistics:** The `PrefixCache` can track and provide statistics
+  on cache hit rates and hit length proportions.
 
 Core Components:
 - `Value`: Dataclass holding the KVCache (`prefix`), token information, and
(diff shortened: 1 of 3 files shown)


jetstream/core/prefix_cache.py
```diff
@@ -863,72 +1091,100 @@ def load(
...
+      value = self._cache.retrieve(
+          stored_key_from_cache, found_layer_idx, device=device
+      )
+      self._hit_statistic.hit(
+          hit_length=common_prefix_length,
+          hit_layer_idx=found_layer_idx,
+          key_length=len(key),
+      )
+      if PRINT_CACHE_STATISTIC:
+        self._print_statistic_without_lock()
```
vipannalla, line 1120: Can you use logger and log_level instead of a new env_var. This is already used in orchestrator -- https://github.com/AI-Hypercomputer/JetStream/blob/main/jetstream/core/orchestrator.py#L109

vipannalla (commented):
Looks good, can you share benchmark results before and with this PR and what metrics did this improve?

mailvijayasingh (approved):

jetstream/core/prefix_cache.py
yuyanpeng-google: Add log_level 11 to prevent too many DEBUG log.

yuyanpeng-google:
> Looks good, can you share benchmark results before and with this PR and what metrics did this improve?

There is no formally result before this PR. There is just some causal small experiments and found that the device_get would block at the critical path. After this PR there is a first version benchmark result in b/397854862

jyj0w0 (approved):

